PR #91, opened 2017-11-18
Fix build under GCC 6.3.0

d99kris:
Build error encountered in `gcc (Ubuntu 6.3.0-12ubuntu2) 6.3.0 20170406` on Ubuntu 17.04:
`nlp/parser/parser.cc: In member function 'void sling::nlp::Parser::Parse(sling::nlp::Document*) const':`
`nlp/parser/parser.cc:291:28: error: 'INFINITY' was not declared in this scope`
`         float max_score = -INFINITY;`
Propose to add include for `cmath` which defines INFINITY.


d99kris:
This has already been fixed.

PR #95, opened 2017-11-22
Replace glog and gflags modules

ringgaard:
This PR replaces the third-party glog and gflags modules with simpler internal implementations that only supports the functionality we need. This simplifies the build process and also makes it easier to include sling as an external library.


base/logging.h
```diff
@@ -15,7 +15,270 @@
...
+
+template <typename T>
+T &&CheckNotNull(const char *file, int line, const char *exprtext, T &&t) {
+  if (t == nullptr) {
+    LogMessageFatal(file, line) << string(exprtext);
+  }
+  return std::forward<T>(t);
+}
+
+#endif
```
terryykoo: Optional: Add "// !defined(LOG)" or something like that

base/logging.h
```diff
@@ -15,7 +15,270 @@
...
-#include "glog/logging.h"
+#include <sstream>
+#include <limits>
+
+#include "base/macros.h"
+#include "base/types.h"
+
+namespace sling {
+
+#ifndef LOG
```
terryykoo: Maybe add a similar #if-guard for the new base/flags.h, in case some piece of code #includes sling/base/flags.h and another flags.h?

ringgaard: Good idea. I have added a guard on DEFINE_VARIABLE.

base/logging.h
ringgaard: Done.

PR #101, opened 2017-11-29
Move C++ code into sling subdirectory

ringgaard:
In order to prevent include file collisions when using SLING in another project, I have moved all the C++ code into a sling subdirectory. This means that all includes will be prefixed with sling, e.g.:

```
#include "sling/frame/store.h"
```

I have also added a SLING_ prefix to all the include guards. This is similar to the setup used by other project (e.g. Tensorflow).


ringgaard:
I did expect some pushback on this, but we are doing this for the greater
good :)

PR #104, opened 2017-12-04
update broken link in readme file 

cloudtrends:
the "sling tensorflow include" link is broken, change it to :
https://github.com/google/sling/blob/master/third_party/tensorflow/include .


cloudtrends:
I signed it!

PR #107, opened 2017-12-07
Python API for SLING frame store

ringgaard:
Python API for using SLING frame stores. This uses the native C++ implementation with Python C API wrappers to access the frames stores.

I have also implemented support for having a wire marker for binary encoded frames. Now, the binary encoding starts with a nul byte to signal that it is binary encoded. Frames in text format will never start with a nul byte. In binary encoding a nul byte is interpreted as a REF(0) tag, which cannot be the first tag in the encoding because there are not reference frames to begin with. The binary wire marker has the advantage that we can distinguish binary and text encoding by just looking at the first byte. This is used by the load and parse methods to automatically detect the encoding. Frames stored in binary encoding without the binary wire marker can be read by forcing binary encoding using the optional binary=True parameter.


ringgaard:
I have created a Python wheel for installing the Python SLING API:

```
sudo pip install http://www.jbox.dk/sling/sling-1.0.0-cp27-none-linux_x86_64.whl
```

sling/pyapi/pyframe.cc
```diff
@@ -0,0 +1,365 @@
...
+  if (op == Py_NE) match = !match;
+  if (match) {
+    Py_RETURN_TRUE;
+  } else {
+    Py_RETURN_FALSE;
+  }
+}
+
+PyObject *PyFrame::GetStore() {
+  Py_INCREF(pystore);
```
rahul1980: Is this refcount decremented when the python object that stores this reference expires?

sling/pyapi/pyframe.cc
```diff
@@ -0,0 +1,365 @@
...
+  GCLock lock(pystore->store);
+  Handle role = pystore->RoleValue(pyname);
+  if (role.IsError()) return nullptr;
+
+  // Get value.
+  Handle value = pystore->Value(pyvalue);
+  if (value.IsError()) return nullptr;
+
+  // Set role value for frame.
+  pystore->store->Set(handle(), role, value);
```
rahul1980: In case the frame has an existing 'role' slot, Store::Set() overwrites the slot value instead of appending.
Perhaps you meant to call Store::Add() instead?

sling/pyapi/pyframe.cc
```diff
@@ -0,0 +1,365 @@
...
+}
+
+PyObject *PyFrame::Find(PyObject *args, PyObject *kw) {
+  // Get role argument.
+  PyObject *pyrole;
+  if (!PyArg_ParseTuple(args, "O", &pyrole)) return nullptr;
+  Handle role = pystore->RoleValue(pyrole);
+  if (role.IsError()) return nullptr;
+
+  // Create iterator for find all slot with the role.
```
rahul1980: find all slot -> finding all slots

sling/pyapi/pyframe.cc
```diff
@@ -0,0 +1,365 @@
...
+      // Create two-tuple for name and value.
+      PyObject *name = pyframe->pystore->PyValue(slot->name);
+      PyObject *value = pyframe->pystore->PyValue(slot->value);
+      return PyTuple_Pack(2, name, value);
+    } else if (role == slot->name) {
+      return pyframe->pystore->PyValue(slot->value);
+    }
+  }
+
+  // More more slots.
```
rahul1980: No more slots.

sling/pyapi/pyarray.cc
```diff
@@ -0,0 +1,200 @@
...
+  // Return array element.
+  return pystore->PyValue(arr->get(index));
+}
+
+int PyArray::SetItem(Py_ssize_t index, PyObject *value) {
+  // Check that array is writable.
+  if (!Writable()) return -1;
+
+  // Check array bounds.
+  if (index < 0 || index >= array()->length()) {
```
rahul1980: For extra coolness, can we also support negative indices like Python :) ?

sling/frame/reader.cc
```diff
@@ -27,6 +27,14 @@ Object Reader::Read() {
   return Object(store_, ReadObject());
 }
 
+Object Reader::ReadAll() {
+  Handle handle;
```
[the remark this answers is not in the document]
ringgaard: Good catch! The input could be empty.

sling/frame/store.cc
```diff
@@ -36,11 +36,11 @@ static const Word kInitialHeap[] = {
   // is frame
   FRAME | NAMED  |  8, 0x18, 0x08, 0x30,
   // id symbol
-  SYMBOL         | 16, 0x20, 0x9fc5c532, 0, 0x38, 0x08,
+  SYMBOL         | 16, 0x20, 0x307A1C66, 0, 0x38, 0x08,
```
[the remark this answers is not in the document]
ringgaard: We switched to the open source version of City hash, and apparently it generates different hash values. This has been hidden by the fact that we have special code for handling these when reading from both text and binary formats. When I tested the Python API though, I noticed that store["id"] did not work.

sling/pyapi/pyarray.cc
ringgaard: Extra coolness added :)

sling/pyapi/pyframe.cc
ringgaard: Yes. In general, when you return a Python object, the caller will have a new reference to it.

sling/pyapi/pyframe.cc
ringgaard: Yes, you're right. It should be Add and not Set. Fixed.

sling/pyapi/pyframe.cc
ringgaard: Done

sling/pyapi/pyframe.cc
ringgaard: Done

sling/pyapi/pystore.cc
```diff
@@ -0,0 +1,558 @@
...
+}
+
+PyObject *PyStore::Load(PyObject *args, PyObject *kw) {
+  // Parse arguments.
+  static const char *kwlist[] = {"file", "binary", nullptr};
+  PyObject *file = nullptr;
+  bool force_binary = false;
+  bool ok = PyArg_ParseTupleAndKeywords(
+                args, kw, "O|b", const_cast<char **>(kwlist),
+                  &file, &force_binary);
```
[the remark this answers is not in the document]
ringgaard: Fixed

sling/pyapi/pystore.cc
```diff
@@ -0,0 +1,558 @@
...
+  bool ok = PyArg_ParseTupleAndKeywords(
+                args, kw, "S|b", const_cast<char **>(kwlist),
+                  &object, &force_binary);
+  if (!ok) return nullptr;
+
+  // Check that store is writable.
+  if (!Writable()) return nullptr;
+
+  // Get data buffer.
+  char *data;
```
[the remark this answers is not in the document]
ringgaard: Yes, it should have been const char *, but the Python C API does not specify const for this parameter. Likewise for PyArg_ParseTupleAndKeywords, where you need to have a const_cast for the kwlist parameter when using the API from C++.

sling/pyapi/pystore.cc
```diff
@@ -0,0 +1,558 @@
...
+
+Handle PyStore::Value(PyObject *object) {
+  if (object == Py_None) {
+    return Handle::nil();
+  } else if (PyObject_TypeCheck(object, &PyFrame::type)) {
+    // Return handle for frame.
+    PyFrame *frame = reinterpret_cast<PyFrame *>(object);
+    if (frame->pystore->store != store &&
+        frame->pystore->store != store->globals()) {
+      PyErr_SetString(PyExc_ValueError, "Frame does not belongs to this store");
```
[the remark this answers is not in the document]
ringgaard: Fixed

PR #109, opened 2017-12-08
Script for making Python wheel for SLING API

ringgaard:
I have made a script for making a [Python Wheel](https://pythonwheels.com/) for installing the SLING Python API. You can build and install it this way:
```
bazel build -c opt sling/pyapi:pysling.so
python tools/build-wheel.py
sudo pip install /tmp/sling-1.0.0-cp27-none-linux_x86_64.whl
```

Bonus: I have also fixed a bug in reading binary frames in Python.


tools/build-wheel.py
```diff
@@ -0,0 +1,126 @@
...
+
+# Wheel package information.
+package = "sling"
+version = "1.0.0"
+dist_dir = package + "-" + version + ".dist-info"
+data_dir = package + "-" + version + ".data/purelib"
+record_filename = dist_dir + "/RECORD"
+
+wheel_dir = "/tmp"
+weel_basename = package + "-" + version + "-cp27-none-linux_x86_64.whl"
```
rahul1980: wheel_basename

ringgaard: Fixed

PR #111, opened 2017-12-11
Python wrapper classes for document, token, and mention frames

ringgaard:
I have also added support for deleting frame slots.


sling/frame/store.cc
```diff
@@ -495,6 +495,39 @@ Handle Store::Extend(Handle frame, Handle name, Handle value) {
   return AllocateHandle(clone);
 }
 
+void Store::Delete(Handle frame, Handle name) {
+  // Get frame.
+  FrameDatum *datum = GetFrame(frame);
+  CHECK(datum->IsFrame());
+
+  // Id slots cannot be deleted.
+  DCHECK(name != Handle::id());
```
rahul1980: why not CHECK instead of DCHECK ?

python/nlp/document.py
```diff
@@ -0,0 +1,239 @@
...
+      for token in self.tokens: array.append(token.frame)
+      self.frame[self.names.document_tokens] = array
+      self.tokens_dirty = False
+
+    # Update mentions in document frame.
+    if self.mentions_dirty:
+      slots = []
+      for mention in self.mentions:
+        slots.append((self.names.document_mention, mention.frame))
+      del self.frame[self.names.document_mention]
```
rahul1980: this only gets rid of the first mention right?
did you mean to do:
self.frame[self.names.document_mention] = None

Similarly below for theme.

ringgaard: The del frame[role] construction calls PyFrame::Assign() which uses Store::Delete(). The Store::Delete() method deletes all slots with a particular name, so I think this is okay.

sling/frame/store.cc
ringgaard: Fixed

PR #114, opened 2017-12-13
Record files

ringgaard:
Record files have records with keys and values. Compression with snappy is supported (enabled by default). I have also implemented a Python API for reading and writing record files:
```
import sling
corpus = sling.RecordReader("local/sempar/dev.rec")
for name, value in corpus:
  print name, len(value)
corpus.close()
```
Record files are much faster than zip files but have almost the same compression ratio, e.g. reading train.zip takes 6000 ms but it only takes 250 ms to read the same content as a record file.

I have also added support for shared frame store and added chaining of frame builder methods.


sling/file/recordio.cc
```diff
@@ -0,0 +1,449 @@
...
+    uint64 chunk_used = position_ % info_.chunk_size;
+    if (chunk_used + size_with_skip > info_.chunk_size) {
+      // Write filler record. For a filler record, the record size includes
+      // the header.
+      Header filler;
+      filler.record_type = FILLER_RECORD;
+      filler.record_size = info_.chunk_size - chunk_used;
+      filler.key_size = 0;
+      int hdrlen = WriteHeader(filler, output_.end());
+      output_.appended(hdrlen);
```
rahul1980: shouldn't there also be a position_ += hdrlen here?

sling/pyapi/pyrecordio.cc
```diff
@@ -0,0 +1,214 @@
...
+  Py_ssize_t pos = PyLong_AsSsize_t(arg);
+  if (pos == -1) return nullptr;
+
+  // Seek to position.
+  if (!CheckIO(reader->Seek(pos))) return nullptr;
+  Py_RETURN_NONE;
+}
+
+PyObject *PyRecordReader::Next() {
+  // Check if the are more records.
```
rahul1980: the -> there

sling/file/recordio.cc
ringgaard: This is somewhat subtle, but as stated in the comment above, the record size for a filler record includes the header.

sling/pyapi/pyrecordio.cc
ringgaard: Fixed

PR #135, opened 2018-02-06
Use recordio format by default

rahul1980:


sling/nlp/document/document-source.cc
```diff
@@ -96,6 +97,55 @@ class ZipDocumentSource : public DocumentSource {
...
+
+  void Rewind() override {
+    if (reader_ != nullptr) {
+      delete reader_;
+      reader_ = new RecordReader(file_);
+    }
+  }
+
+ private:
+  RecordReader *reader_ = nullptr;
```
ringgaard: You could make this statically allocated since you are initializing it in the constructor.

rahul1980: I tried that earlier, but the problem is in DocumentSource::Rewind(). In that method, simply calling RecordReader::Seek(0) doesn't work (it gives a corrupt header error during subsequent calls to ReadRecord()). Basically on rewinding I need to do most of the work that is currently done in the constructor of RecordReader, hence I am creating it again.

PR #139, opened 2018-02-16
Map-reduce task components

ringgaard:
Task components for doing in-process map-reductions. 

* Mapper - Thin wrapper on top of the task interface to implement map functionality.
* Sharder - Shards messages according to key fingerprint and outputs to a set of channels.
* Sorter - Collects all incoming messages and outputs them sorted by key.
* Reducer - Reducer component for collecting all messages with the same key.
* Accumulator - Combines key counts using hash structure. This also has a sum-reducer for processing aggregated counts.

This PR also has a bunch of small utility components:

* Identity - Identity mapper and reducer.
* Message printer - Outputs messages to log.
* Null sink - task for consuming and discarding messages.


sling/task/mapper.h
```diff
@@ -0,0 +1,71 @@
...
+
+  // Value for message.
+  Slice value() const { return value_; }
+
+ private:
+  Slice key_;
+  Slice value_;
+};
+
+// A mapper process all the input message in the Map() method and can output.
```
rahul1980: process -> processes,  remove trailing period.

ringgaard: Fixed

PR #146, opened 2018-02-27
Bug fix in DocumentFeatures

rahul1980:
DocumentFeatures is always resized when we moved to the next document, and so the old features from the previous document are not reset. This is fine if the new feature value is guaranteed to be recomputed (e.g. for punctuation), but that's not the case for Hyphen and Quote features. This fix just initializes those features to NO_HYPHEN and NO_QUOTE.


rahul1980:
I will push this commit to the caspar branch as well (no need to make another PR, I think).

PR #148, opened 2018-02-27
Fix return value in RecordIODocumentSource::Next

entn-at:
`RecordIODocumentSource::Next` returned `false`, but the return type is `Document *`, resulting in a compilation error. The overridden base class method (`DocumentSource::Next`) also returns `nullptr`.


entn-at:
I signed it!

PR #150, opened 2018-03-05
Omit role features if there are no roles

rahul1980:


rahul1980:
Submitting this PR now. Will take a look at the exporter and runtime next.

PR #153, opened 2018-03-05
Skip calling RoleGraph::Compute() if there are no role features

rahul1980:


sling/nlp/parser/parser.cc
```diff
@@ -525,7 +525,7 @@ void ParserInstance::ExtractFeaturesFF(int step) {
   }
 
   // Extract role features.
-  if (parser_->frame_limit_ > 0) {
+  if ((parser_->frame_limit_ > 0) && (parser_->roles_.size() > 0)) {
```
ringgaard: There is no need for the extra parentheses, so you can change this to:

if (parser_->frame_limit_ > 0 && !parser_->roles_.empty())

PR #154, opened 2018-03-06
Wikipedia parsing

ringgaard:
Components for parsing Wikipedia articles in Wiki markup format and extract the plain text from the articles. This will also resolve all redirects and convert all ids to Wikidata ids. This also outputs the name statistics based on titles, redirects and disambiguations.


sling/nlp/wiki/wikipedia-documents.cc
```diff
@@ -0,0 +1,565 @@
...
+                mention.AddLink(n_phrase_evokes_, link);
+                article.Add(n_document_mention_, mention.Create());
+              }
+              num_anchors_->Increment();
+            }
+          }
+          break;
+        }
+
+        default: ;
```
rahul1980: default: break;

sling/nlp/wiki/wikipedia-map.cc
```diff
@@ -0,0 +1,167 @@
...
+  // Initialize page type mapping.
+  typemap_[store_.Lookup("/w/item/kind/article")] = ARTICLE;
+  typemap_[store_.Lookup("/w/item/kind/disambiguation")] = DISAMBIGUATION;
+  typemap_[store_.Lookup("/w/item/kind/category")] = CATEGORY;
+  typemap_[store_.Lookup("/w/item/kind/list")] = LIST;
+  typemap_[store_.Lookup("/w/item/kind/template")] = TEMPLATE;
+  typemap_[store_.Lookup("/w/item/kind/infobox")] = INFOBOX;
+
+  // Allow duplicate symbols.
+  options_.symbol_rebinding = true;
```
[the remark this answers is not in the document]
ringgaard: I forgot why I added this :( I have added a TODO comment for myself to investigate is this is really needed, and if so, document why this is the case.

sling/nlp/wiki/wikipedia-documents.cc
ringgaard: Fixed.

PR #156, opened 2018-03-19
Name and phrase tables

ringgaard:
Build name and phrase tables for knowledge base. These are store in repository files using pointer-free structures that can be loaded directly into memory.

The profile alias reducer selects aliases for entities using an algorithm that is similar to wikifuse.

The name stable contains the names in a sorted list to allow looking up names based in name prefix. This is used for the incremental search in the knowledge browser.

The phrase table stores the names in a hash tables as fingerprints, and is useful for looking up phrases in a tokenized document.


sling/nlp/wiki/name-table.cc
```diff
@@ -0,0 +1,101 @@
...
+  std::vector<std::pair<uint32, const EntityItem *>> matching_entities;
+  for (auto it : entities) {
+    matching_entities.emplace_back(it.second, it.first);
+  }
+  std::sort(matching_entities.rbegin(), matching_entities.rend());
+
+  // Copy matching entity ids to output.
+  matches->clear();
+  for (const auto &item : matching_entities) {
+    if (matches->size() >= limit) break;
```
rahul1980: we don't need this check since matching_entities.size() <= limit  (enforced on line 64)

ringgaard: Good point. Fixed.

sling/file/repository.cc
```diff
@@ -0,0 +1,296 @@
...
+char *Repository::GetMutableBlock(const string &name) {
+  for (Block &block : blocks_) {
+    if (block.name == name) return block.data;
+  }
+  return nullptr;
+}
+
+size_t Repository::GetBlockSize(const string &name) const {
+  for (const Block &block : blocks_) {
+    if (block.name == name) return block.size;
```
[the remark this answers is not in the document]
ringgaard: Yes, this way you can also use this method on blocks backed by a file.

PR #164, opened 2018-04-11
Python API for task system

ringgaard:
Python interface to the SLING task system.  This extends the pysling module with methods for accessing the task system and the task dashboard. This interface is used for making a workflow abstraction in Python to define a set of tasks to be run and connecting these together.  The PR also contains methods for setting up the tasks for extracting data from Wikipedia and Wikidata. The run.py program is a command line interface to the task system which allows you to run the wiki tasks from the command line.

This PR also add a Python interface to the phrase table code.


python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+        r = self.resource_map.get(key)
+        if r == None:
+          r = Resource(filenames[shard], Shard(shard, n), format)
+          self.resource_map[key] = r
+          self.resources.append(r)
+          resources.append(r)
+      return resources
+
+  def channel(self, producer, name="output", shards=None, format=None):
+    """Adds one or more channels to wrofklow. The channel(s) are connected as
```
rahul1980: typo: workflow

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+        writer = self.task(tasktype, name=name, shard=Shard(shard, fanout))
+      writer.attach_output("output", output[shard])
+      writer_tasks.append(writer)
+
+    # Connect producer(s) to writer task(s).
+    if isinstance(input, list) and len(input) == 1: input = input[0]
+    if fanout == 1: writer_tasks = writer_tasks[0]
+    self.connect(input, writer_tasks)
+
+    return output
```
rahul1980: did you intend to return writer_tasks here? 'output' is already an input of this method.

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+        channels.append(self.read(arg))
+      elif isinstance(arg, list):
+        for elem in arg:
+          if isinstance(elem, Channel):
+            channels.append(elem)
+          elif isinstance(elem, Resource):
+            channels.append(self.read(elem))
+          else:
+            raise Exception("illegal element")
+        pass
```
rahul1980: this can be removed

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+      self.connect(pipes, sorters)
+
+    # Return output channel from sorters.
+    outputs = self.channel(sorters, format=format_of(input))
+    return outputs
+
+  def reduce(self, input, output, type=None, params=None, name=None):
+    """Reduce input and write reduced output."""
+    if type == None:
+      # No reducer (i.e. i.e. identity reducer), just write input.
```
rahul1980: typo: duplicate "i.e."

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+
+    # Shuffling of map output.
+    shuffle = self.shuffle(mapping, shards=shards)
+
+    # Reduction of shuffled map output.
+    self.reduce(shuffle, output, reducer, params=params)
+    return output
+
+  def start(self):
+    """Start workflow. The workflow with be run in the background, and the
```
rahul1980: typo: with -> will

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+    if self.job == None: return True
+    if timeout != None:
+      return self.job.wait_for(timeout)
+    else:
+      self.job.wait()
+      return True
+
+  def done(self):
+    """Check if workflow is done."""
+    if self.job == None: return Trye
```
rahul1980: Trye -> True

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+def start_monitor(port):
+  """Start task monitor."""
+  api.start_task_monitor(port)
+
+def stop_monitor():
+  """Stop task monitor."""
+  global active
+  if active:
+    log.info("sending final status to monitor")
+    api.finalize_dashboard();
```
rahul1980: should we also set active = False here?
Also, there is a trailing semi-colon here.

sling/pyapi/pytask.cc
```diff
@@ -0,0 +1,316 @@
...
+  for (int i = 0; i < PyList_Size(resources); ++i) {
+    PyObject *pyresource = PyList_GetItem(resources, i);
+    PyObject *pyformat = PyAttr(pyresource, "format");
+    PyObject *pyshard = PyAttr(pyresource, "shard");
+
+    const char *name = PyStrAttr(pyresource, "name");
+    Format format = PyGetFormat(pyformat);
+    Shard shard = PyGetShard(pyshard);
+
+
```
rahul1980: extra empty line

python/task/corpora.py
```diff
@@ -0,0 +1,114 @@
...
+             default="en",
+             metavar="LANG")
+
+flags.define("--languages",
+             help="list of languages to process",
+             metavar="LANG,...")
+
+flags.define("--wikidata",
+             help="wikidata version",
+             #default="20161031",
```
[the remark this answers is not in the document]
ringgaard: I included the old version to make it easy to go back to older versions to test for regressions, but I have removed the 20161031 version now that the 2018 version has been tested.

python/task/corpora.py
```diff
@@ -0,0 +1,114 @@
...
+    arg.languages = arg.language
+  if arg.languages == "ALL":
+    arg.languages = "en,da,sv,no,de,fr,es,it,nl,pt,pl,fi"
+  arg.languages = arg.languages.split(",")
+
+flags.hook(post_process_flags)
+
+def wikidata_url():
+  """URL for downloading Wikidata dump."""
+  return "https://dumps.wikimedia.org/wikidatawiki/entities/" + \
```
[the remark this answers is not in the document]
ringgaard: I think you can probably set the "wiki date" to be latest, but I will not encourage using the latest versions since this makes version tracking faster. The idea is that we update the wikipedia and wikidata version flags to the current version when we have tested that this version works.

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+  def __repr__(self):
+    s = "Resource(" + self.name
+    if self.shard != None: s += str(self.shard)
+    if self.format != None: s += " as " + str(self.format)
+    s += ")"
+    return s
+
+
+class Binding:
+  """A binding binds a resouce to a task input or output."""
```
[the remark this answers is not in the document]
ringgaard: Fixed

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+
+  def __repr__(self):
+    s = str(self.task)
+    s += "." + self.name
+    if self.shard != None: s += str(self.shard)
+    return s
+
+
+class Channel:
+  """A channel is used for seding messages from a producer task to a consumer
```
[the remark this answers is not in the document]
ringgaard: Fixed

python/task/workflow.py
```diff
@@ -0,0 +1,703 @@
...
+        self.add_param(name, value)
+
+  def __repr__(self):
+    s = self.name
+    if self.shard != None: s += str(self.shard)
+    return s
+
+
+class Scope:
+  """A score is used for defined a name space for task names. It implements
```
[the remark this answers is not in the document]
ringgaard: Fixed

python/task/workflow.py
ringgaard: Fixed

python/task/workflow.py
ringgaard: I think this is what I intended. I agree that returning one of the input arguments is not really providing any new information to the caller, but I think it is more convenient for the caller to get the list of output resource rather than the list of writer tasks, although this remains to be seen since we are not really using this anywhere.

python/task/workflow.py
ringgaard: Done

python/task/workflow.py
ringgaard: Fixed. However, the comment is slightly misleading since we don't actually use an identity reducer in this case. The task system is more flexible that a standard mapreduce, where you are forced to provide both a mapper and a reducer.

python/task/workflow.py
ringgaard: Fixed

python/task/workflow.py
ringgaard: Fixed

python/task/workflow.py
ringgaard: The global active flag just tracks if the workflow system has been activated, not whether it is still active. I added this so that python/run.py --help does not try to stop the monitor and write a workflow log.

Semi-colon removed.

sling/pyapi/pytask.cc
ringgaard: Fixed.

PR #174, opened 2018-05-03
Lexical encoder

ringgaard:
The lexical encoder component is a neural network component for encoding sentences in a document. It is a neural network component that has a feature extractor for extracting features from sentences. These features are mapped though embeddings. The word embeddings can be initialized from pre-trained embeddings. On top of the feature encoder is has a bi-directional LSTM which outputs the hidden activations for each token. The lexical encoder can be used at both training and decoding time.


sling/nlp/document/lexical-encoder.cc
```diff
@@ -0,0 +1,398 @@
...
+
+  // Extract lexical features from document.
+  DocumentFeatures features(&lex_.lexicon_);
+  features.Extract(document, begin, end);
+
+  // Compute feature vector for all tokens in range.
+  int length = end - begin;
+  fv_.resize(length);
+  for (int i = 0; i < length; ++i) {
+    auto *e = new LexicalFeatureExtractor(lex_);
```
rahul1980: This repeated allocation of LexicalFeatureExtractors is only for gradient propagation, right? IIRC, for inference, in theory we can reuse a single LexicalFeatureExtractor as long as we zero-out stale feature ids.

sling/nlp/document/features.cc
```diff
@@ -21,14 +21,16 @@
...
+  int length = end - begin;
+  features_.resize(length);
   bool extract_prefixes = lexicon_->prefixes().size() != 0;
   bool extract_suffixes = lexicon_->suffixes().size() != 0;
   int oov = lexicon_->oov();
   bool in_quote = false;
-  for (int i = 0; i < document.num_tokens(); ++i) {
-    string word = document.token(i).text();
+  for (int i = 0; i < length; ++i) {
+    string word = document.token(begin + i).text();
```
rahul1980: As per our discussion, let's go with the following policy. This is best implemented inside the Lexicon class.

LEXICON BUILDING
-----------------------------
Given a word w,
1. w = NormalizeDigits(w) (if normalize_digits = true). No need for lowercasing.
2. Add (w, count=1) to the word vocab.
3. Compute prefixes and suffixes of w. Note w is the normalized version of the original word.
4. Save these affixes to the corresponding affix tables.
5. Optionally compute the other shape features of w and save them in the lexicon, so we don't have to compute them again and again at runtime.

RUNTIME
--------------
Given word w,
1. w = NormalizeDigits(w)
2. Lookup the lexicon for w
3. If found, return all the precomputed features for w
4. If not found:
    a) Return OOV for word feature
    b) Compute longest affixes of w for prefix/suffix features, and return their ids.
    c) If no such ids are in the affix tables, return OOVs. This means that suffix and prefix features will now 
        have OOVs
    d) Compute shape features on the fly for w and return their values. OOV doesn't arise here.

This has two main differences than the existing implementation:
- Everything in this policy is based off the digit-normalized version.
- Affix features have an OOV.
- (Optional) We can precompute and store the shape features in the lexicon.

ringgaard: I think we should support case-insensitive vocabularies. This will make the model smaller and we have the back-off features to account for case. I will start working on making the necessary changes to the lexicon and the feature extractor for this and send you a new PR with these changes. 

I agree that we should cache the features for the know words. If we have a case-insensitive lexicon, we only need to compute the word capitalization feature. The rest are invariant to normalization.

We could use the empty affix as a back-off.  We are already backing-off to the longest known affix now, but the empty affix will ensure that we always generate at least one feature for the prefixes and the suffixes.

sling/nlp/document/lexical-encoder.cc
ringgaard: Yes, this is only needed during learning. We could consider reusing these so we just keep extending the extractors_ array instead of reallocating this for each sentence.

sling/util/random.h
```diff
@@ -0,0 +1,58 @@
...
+    return dist_(prng_);
+  }
+
+  // Return uniformly distributed random number r=p*scale+bias, 0<=p<1.
+  float UniformFloat(float scale, float bias) {
+    return dist_(prng_) * scale + bias;
+  }
+
+  // Return uniformly distributed random number between 0 and n (exclusive).
+  float UniformInt(int n) {
```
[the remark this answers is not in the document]
ringgaard: Oops. Fixed.

PR #181, opened 2018-05-15
OS-specific build targets

ringgaard:
I have added support for OS-specific build targets so build rules can distinguish between Linux and OSX. I have fixed a number of build issues under OSX so the parser tools and the pysling components can now be build under OSX.

OSX does not support ELF files (it uses Mach-O files instead) so the ELF writer cannot be used under OSX. This also means that we cannot use the embed-data rules for embedding web content in the binaries. Also, OSX does not support epoll, so the HTTP server cannot build under OSX.

This means that the task dashboard cannot run under OSX, so when we build under OSX we disable the dashboard.

PS: Although this PR does improve OSX support under SLING, OSX support is still experimental.


ringgaard:
After some internal discussion, we have decided to NOT support OSX. We are only using SLING under Linux ourselves, and since we are not testing SLING under OSX in any way, we would probably keep breaking the OSX build. However, we will try to make sure that SLING can be built with both GCC and clang.

I have updated this PR to remove the OSX-specific changes, but I have kept the parts that improves standard compliance, especially with respect to integer types.

I also removed the learning rate update code that accidentally made into into the PR.

PR #183, opened 2018-05-16
AVX512 support for JIT assembler

ringgaard:
The n1-highcpu-96 machines in GCE supports AVX512, so I have added support for the AVX512 instruction set to the JIT assembler. 

AVX512 extends the SIMD registers to 512 bit and there are now 32 of these called zmm0...zmm31. The AVX512 instructions use a new EVEX encoding which supports 32 SIMD registers as well as a number of new features like broadcasting, static rounding, exception suppression.

The AVX512 instruction set also supports masking using special masking registers, k0...k7, which is useful when handling residual code in tensor ops.

AVX512 add many new instructions, including extensions to existing ones, so I have made a tool, `opt2hdr.py`,  that can generate the all instruction methods for the assembler the instruction descriptions from the Intel manuals, e.g.:

```
EVEX.NDS.128.0F.W0 58 /r VADDPS xmm1 {k1}{z}, xmm2, xmm3/m128/m32bcst	C	V/V	AVX512VL AVX512F	Add packed single-precision floating-point values from xmm3/m128/m32bcst to xmm2 and store result in xmm1 with writemask k1.
EVEX.NDS.256.0F.W0 58 /r VADDPS ymm1 {k1}{z}, ymm2, ymm3/m256/m32bcst	C	V/V	AVX512VL AVX512F	Add packed single-precision floating-point values from ymm3/m256/m32bcst to ymm2 and store result in ymm1 with writemask k1.
EVEX.NDS.512.0F.W0 58 /r VADDPS zmm1 {k1}{z}, zmm2, zmm3/m512/m32bcst{er}	C	V/V	AVX512F	Add packed single-precision floating-point values from zmm3/m512/m32bcst to zmm2 and store result in zmm1 with writemask k1.
``` 

This is converted into this code snippet in avx512.h:

```
void vaddps(ZMMRegister dst, ZMMRegister src1, ZMMRegister src2, Mask mask = nomask, RoundingMode er = noround) {
  zinstr(0x58, dst, src1, src2, 0, mask, EVEX_BT4 | EVEX_ENDS | EVEX_L128 | EVEX_L256 | EVEX_L512 | EVEX_M0F | EVEX_W0 | evex_round(er));
}
void vaddps(ZMMRegister dst, ZMMRegister src1, const Operand &src2, Mask mask = nomask) {
  zinstr(0x58, dst, src1, src2, 0, mask, EVEX_BCST | EVEX_BT4 | EVEX_ENDS | EVEX_L128 | EVEX_L256 | EVEX_L512 | EVEX_M0F | EVEX_W0);
}
```


third_party/jit/cpu.cc
```diff
@@ -315,11 +313,31 @@ void CPU::Initialize() {
...
 
+#if defined( __linux__)
+#include <sys/sysinfo.h>
+
 int CPU::Processors() {
-  Probe();
   return get_nprocs();
 }
 
+#elif defined(__APPLE__)
```
rahul1980: do we still need this and the other header guards?

ringgaard: The get_nprocs(3) function is a "GNU extension" whereas the rest of the JIT assembler code is POSIX compliant as far as I can tell, so this is to avoid making the code Linux-specific. We could drop the `#ifdef __APPLE__` part, but we would need to keep the `#else` part for the code to compile on all POSIX platforms.

PR #184, opened 2018-05-16
Update Python wheel with new myelin and task components

ringgaard:
I have updated the Python wheel to include the new myelin and task components. I have moved the myelin components from `sling/nlp/parser/trainer/flow` to `python/myelin` and removed the old `flow.py` module that is no longer needed after we removed the TF trainer.

I have updated the version number for the sling wheel to 2.0.0 to distinguish the caspar version from the sempar version.


sling/nlp/parser/trainer/pytorch_modules.py
```diff
@@ -23,11 +23,9 @@
 sys.path.insert(0, "sling/nlp/parser/trainer")
 from parser_state import ParserState
 
-sys.path.insert(0, "sling/nlp/parser/trainer/flow")
-
-import nn as flownn
-import flow as flow
-import builder as builder
+import sling.myelin.nn as flownn
```
rahul1980: does this work without any other changes? e.g. do we also need to change python/__init__.py?
Note that the new path of nn.py is python/myelin/ and not sling/myelin.

ringgaard: I think this will work although I have not tested it (shame on me). When you run the parser trainer, we assume that you have pysling installed (or linked), so the Python package directory will contain a `sling` subdirectory that will contain `myelin/nn.py`.

The declarations in `python/__init__.py` are only needed to alias classes into the top sling namespace, e.g. the Document class is actual sling.nlp.document.Document, but we alias this so you can use this as sling.Document. We can consider doing this for some of the myelin classes but we are not relying on that at the moment.

PR #201, opened 2018-06-20
Text normalization

ringgaard:
I have implemented support for consistent and configurable text/phrase/token normalization throughout SLING. The Unicode::Normalize() function now supports the following types of normalization:

* CASE - lower-case letters
* LETTER - remove diacritics from letters
* DIGIT - replace all digits with 9
* PUNCTUATION - remove punctuation
* WHITESPACE - remove whitespace

Each of the normalizations can be selected individually or in combination.

The affix table now contains the empty affix as a backoff feature. Also, the affixes are now computed on the normalized version of the token text.

The lexicon now supports caching of lexical features. It uses the normalization flags to determine which featured that can be cached and which features need to be re-computed on the fly, e.g. in a case-insensitive lexicon, the upper and lower case features need to be recomputed. The document feature extractor will also recompute context-sensitive features.

The Python lexical encoder now writes the normalization used in the trainer to the lexicon blob, which is used by the C++ lexical encoder to ensure that the runtime uses the same normalization as the trainer.

Token fingerprints are now computed lazily in the Document class.

The Fingerprinter class now supports configurable text normalization.

The PhraseTokenizer now supports configurable normalization for computing finger prints.

The name and phrase tables now store the text normalization flags in the repository so we can ensure that the same normalization is applied on look-ups.

Bonus:

- The lexical encoder now supports pre-trained word embeddings with a smaller dimension than the encoder uses. The remaining dimensions are zeroed in that case.

- Fix bug in command line parsing for boolean flags.

- Small optimization in frame symbol lookup when there are no symbols in store.


python/myelin/lexical_encoder.py
```diff
@@ -25,7 +25,9 @@ def __init__(self, flow,  spec, lstm_feature_embeddings, lr_lstm, rl_lstm):
     lexicon.type = "dict"
     lexicon.add_attr("delimiter", 10)
     lexicon.add_attr("oov", spec.words.oov_index)
-    lexicon.add_attr("normalize_digits", spec.words.normalize_digits)
+    normalization = ""
+    if (spec.words.normalize_digits) normalization = "d"
+    lexicon.add_attr("normalization", normalization)
```
rahul1980: As discussed over chat:
On the next line, we set lexicon.data to str(spec.words), which uses unnormalized form of words. Later in Lexicon::InitWords(), we expect normalized words, so this would cause a disparity.

sling/nlp/parser/trainer/spec.py
```diff
@@ -298,7 +298,7 @@ def build(self, commons, corpora):
     # For compatibility with DRAGNN, suffixes don't have an OOV item.
     self.commons = commons
     self.words = Lexicon(self.words_normalize_digits)
-    self.suffix = Lexicon(normalize_digits=False, oov_item=None)
+    self.suffix = Lexicon(self.words_normalize_digits, oov_item=None)
```
rahul1980: Another TODO (for me): Make sure that the suffix feature here is the same as the new implementation.

python/myelin/lexical_encoder.py
ringgaard: The Lexicon.add() method normalizes the words when they are inserted, so I believe that the Lexicon.index_to_item contains the normalized words. Am I missing something here?

sling/nlp/parser/trainer/spec.py
ringgaard: The C++ affix table still supports the case where we do not have the empty affix, but I think we could simplify it a bit if we could always assume that it was there.

PR #205, opened 2018-06-29
Run tasks written in Python

ringgaard:
The task system can now run tasks written in Python.

The task system now releases the Python global interpreter lock (GIL) when running workflow tasks to allow execution of Python-based tasks in other threads.

A Python-based task is defined by a class with a run(task) method. The task can get the inputs, outputs, and parameters for the task, as well as updating up performance counters.

Python-based tasks can be registered with the sling.task.workflow.register_task method.

The components registry has been updated to allow registration of lambdas as component factories.

Wikidata and Wikipedia download are now run as tasks under the task system which allows these downloads to run in parallel.


python/task/download.py
```diff
@@ -0,0 +1,111 @@
...
+    if wf == None: wf = Workflow(name)
+    self.wf = wf
+
+  #---------------------------------------------------------------------------
+  # Wikipedia dumps
+  #---------------------------------------------------------------------------
+
+  def wikipedia_dump(self, language=None):
+    """Resource for wikipedia dump. This can be downloaded from wikimedia.org
+    and contrains a full dump of Wikipedia in a particular language. This is
```
rahul1980: typo: contains

ringgaard: Fixed

PR #217, opened 2018-08-06
General SIMD matrix multiplication kernel

ringgaard:
Due to the differentiation rule for matrix multiplications, the gradient functions end up having all sorts of combinations of transposed inputs and outputs for the MatMuls. In order to support this, I have implemented a general MatMul kernel that supports all combinations of tensor element orders as well as all combinations of transpose of the inputs and output.

It uses a SIMD assembler, which has code generators for all the different instruction set variants (vector and scalar sse, avx128, avx256, avx512) for generating the vector operations needed for vectorized matrix multiplication. The SIMD assembler constructs a cascade of vector code generators based on the vector size and the instruction sets supported by the CPU. From this, a strategy for unrolling, repetition, and residuals (with masking if supported) is computed. The SIMD  assembler makes it simpler to implement general kernels that support all the instruction sets for different element types.

The SIMD matrix multiplication kernel selects one of four algorithms for implementing the MalMul based on the layout and transpose of the input and output tensors:

1) Non-strided vertical summation (RR)
2) Strided vertical summation (CR)
3) Horizontal summation (RC)
4) Non-vectorized columnwise (CC)

The SIMD matmul kernel handles both aligned and unaligned loads as well as row padding. The SIMD matmul kernel also supports output accumulation (AssignAddMatMul).

Bonus:

- More compact register allocation in expressions by releasing source registers before allocating destination register.
- Support for using identity ops for broadcasting.
- Added name to transformation components to allow for better tracing of transformations.
- Added check for unique names in flow consistency checker.
- Pruning of unused ops from gradient function.
- Gradient functions for sqrt, sum, and transpose.
- Revamped flow to dot graph converter.


ringgaard:
Thanks for the review.

PR #218, opened 2018-08-07
Predicates and conditionals in expression generator

ringgaard:
I have added support for predicate variables in the expression generator. The predicates are used for implementing the comparison operators (Equal, NotEqual, Less, LessEqual, Greater, GreaterEqual) as well as boolean functions (And, Or, Xor, Not, AndNot). The expressions now also supports conditionals like Cond(p,x,y) and Select(p,x).

A predicate variable can only have two values: zero or all ones. The comparison operators now output predicate variables and the standard boolean operators (and, or, not, xor) now operates on predicates. The Cond(p,x,y) operator can select between two values based on the value of the predicate, and Select(p,x) is zero if the predicate is 0 (i.e. false) and x if the predicate is true.

The predicates are implemented as bit masks for the vector instruction generators except avx512, which uses opmask registers (k1-k7) for implementing predicate logic and the CUDA expression generator which uses the CUDA pred type for implementing predicate operations.

The following optimization transformations have been implemented:
  - Not(Not(x)) = x
  - Not(Equal(x)) = NotEqual(x)
  - Not(NotEqual(x)) = Equal(x)
  - Not(Less(x)) = GreaterEqual(x)
  - Not(LessEqual(x)) = Greater(x)
  - Not(Greater(x)) = LessEqual(x)
  - Not(GreaterEqual(x)) = Less(x)
  - And(x,Not(y)) = AndNot(x,y)
  - And(Not(x),y) = AndNot(y,x)
  
Exp, Log, Tanh, and ReluGrad are now implemented using predicates and conditionals.

Bonus:

SHR23 and SHL23 have been replaced with CVTEXPINT and CVTINTEXP.

Myelin Compiler class to wrap the flow analysis and network compilation steps
as well as handing diagnostics.


ringgaard:
Thanks for the review. I have fixed the typo.

  -- Michael

PR #219, opened 2018-08-07
Make suffix features identical across LexicalEncoder and Python trainer

rahul1980:
- The Python trainer didn't have the empty affix, so it has been added there.
- Since the empty affix is present by default in AffixTable (cf. Reset()), we shouldn't add it again while reading an AffixTable. 
- The python trainer now mimics the affix computation logic in the parser, viz. given a normalized word W, a max affix length L, and an affix table T:
a) Get the longest affix A for W in T. If none exists, return the empty affix as the sole result.
b) Return max(len(A), L) suffixes of A
c) If len(A) < L, additionally return the empty affix.

### DIFF
--- a/sling/nlp/document/affix.cc
+++ b/sling/nlp/document/affix.cc
@@ -153,7 +153,7 @@ Affix *AffixTable::AddAffixesForWord(Text word) {
   // Try to find successively shorter affixes.
   Affix *top = nullptr;
   Affix *ancestor = nullptr;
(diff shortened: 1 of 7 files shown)


rahul1980:
I have modified the suffix feature to always return the empty affix. This means that we can now have 1+max_length outputs from this feature, so I have modified the flow part of the code accordingly. I have also done some minor cleanup.

PR #224, opened 2018-08-10
Utilities for training data preparation

rahul1980:
- Script/Method to build a commons store from given recordio(s) of documents.
- Script/Method to check a recordio against common errors.
- Save actions table and cascade specification in the common store instead of separate blobs/stores.
- Trainer now creates the commons store if one is not provided (using the script in the first bullet).
- A few short-hands in sling.Document and sling.Mention classes.
- Extend store.frame(x) method to create a frame with id x if x is a string. This is a shorthand for the clunkier store.frame({"id": x}) method.


sling/pyapi/pystore.cc
```diff
@@ -260,11 +260,17 @@ PyObject *PyStore::Symbols() {
...
-
-  // Parse data into slot list.
+  
   GCLock lock(store);
   std::vector<Slot> slots;
-  if (!SlotList(arg, &slots)) return nullptr;
+
+  // If the argument is a string, create a frame with that id.
+  if (PyString_Check(arg)) {
+    slots.emplace_back(Handle::id(), SymbolValue(arg));
```
ringgaard: What about just using store['myid'] to create an empty frame? Then you don't need any special support in NewFrame. This will also take care of the proxy flag.

rahul1980:
Thanks Michael, I have updated the documentation, so PTAL.

sling/pyapi/pystore.cc
rahul1980: As discussed earlier, store["myid"] creates an unbound symbol that is not saved, e.g. during Encode().

ringgaard:
LGTM

PR #234, opened 2018-08-22
Python API for Wikidata converter

ringgaard:
I have moved the Wikidata converter code to a separate module, so we can use it from the Python API. The is needed for the sling wiki bot when checking the revision history.


sling/nlp/wiki/wiki.h
```diff
@@ -24,6 +24,13 @@
 namespace sling {
 namespace nlp {
 
+// Wikidata object types.
+enum WikidataType {
+  WIKIDATA_ITEM = 0,
+  WIKIDATA_PROPERTY = 1,
+  WIKIDATA_LEXEME = 1,
```
anders-sandholm: Suppose WIKIDATA_LEXEME should be equal to 2.

ringgaard: Oops, that's a bug. Fixed. I did not notice this during my test because my Wikidata dump does not have any lexemes.

PR #235, opened 2018-08-22
Task for running multi-threaded learning

ringgaard:
I have implemented a task class for running learning using a pool of worker threads and periodically invoking evaluation.

I have also added support for queued input to process tasks, so data can be pulled instead of pushed.

Extras:
- Momentum optimizer.
- Random initialization for variable
- A little const-craziness in myelin
- Methods for locating corresponding gradient functions and variables.


sling/task/learner.cc
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    // Run evaluation.
+    Evaluate(epoch_, model);
+
+    // Check if we are done.
+    if (epoch_ >= epochs_) break;
+  }
+
+  // Wait until workers completes.
```
rahul1980: Is 1 epoch == 1 mini-batch?

Say we have 10 worker threads, each processing their own mini-batches. Say the first 5 finish early, and that brings the total number of mini-batches (=epochs) to epochs_. So the main thread would run evaluation, but the other 5 threads are still busy updating the network parameters. So the eval might be slightly off. Further,  the parameters we save at the end are also slightly different from the ones used for evaluation (since other threads might update the params further during pool.Join()). I just want to ensure whether this minor issue is expected, and ignored for the sake of implementation simplicity.

ringgaard: The evals are running while the parameters are being updated in parallel, but this is by design. However, I have added a final eval after all the workers complete, so the final eval will always be done on the final model. I have also implemented a 1 sec polling interval for the ramp period to prevent these from stalling completion of training.

PR #242, opened 2018-08-30
Record file index

ringgaard:
Record files can now have an internal index that allows for fast lookup of records by key.

The `RecordDatabase` class can be used for looking up records in a sharded set of record files, like the file sets from the wiki processing pipeline.

I have added support for indexed record files to the Python API, e.g.:
```
# Write records to indexed record file.
N=1000
writer = sling.RecordWriter("/tmp/test.rec", index=True)
for i in range(N):
  writer.write(str(i), "Data for record number " + str(i))
writer.close()

# Look up each record in record database.
db = sling.RecordDatabase("/tmp/test.rec")
for i in range(N):
  print db.lookup(str(i))
db.close()
```

A record file without an index can be reindexed using the `index` tool.


ringgaard:
Thanks for the review.

PR #243, opened 2018-09-03
Calendar bug fixes

ringgaard:
* Fix bug for initializing Date object from item with qualified P585 slots.
* Prevent crash when initializing Calendar without calendar schema definitions.
* Fix truncation for date strings with signs.


ringgaard:
Thanks for the speedy review.

PR #244, opened 2018-09-05
Memory-mapped repository files

ringgaard:
Blocks in repository files can now be memory-mapped. This speeds up loading large repositories like the phrase table. We can stil fall back on reading blocks from files in case the file system does not support memory mapping or if the blocks are not page aligned.

The File API has been extended to support memory mapping.

The items are now resolved lazily in the phrase table.

The File::Read() method is now ensuring full reads, even from file handles returning partial data.

We can now initialize a phrase table in under one second.


ringgaard:
Thanks for the review.

PR #245, opened 2018-09-05
Add optional oov input to Gather

ringgaard:


ringgaard:
Thanks for the review. Please let me know if this works for the linked features.

  -- Michael

PR #246, opened 2018-09-06
ISO 8601 dates

ringgaard:
I have added an iso() method to the Python Date class that returns the date in ISO 8601 format.


ringgaard:
Thanks for the speedy review.

PR #248, opened 2018-09-07
Frame store snapshots

ringgaard:
Although frame store loading is already pretty fast (8 GB in 30 secs), faster loading of the knowledge base store is useful when prototyping Python code.

I have added support for creating frame store snapshots. The snapshot files contain a dump of the store heaps that can be loaded directly into a store without any decoding. You can use the `snaps` tool for creating a snapshot of a frame store file. The snapshot file has the same filename as the underlying store file with `.snap` appended.

When a store is loaded, we now check if there is a valid snapshot file for the store, and use this if the snapshot is newer than the underlying store file. The snapshot is only used when loading into a pristine store, i.e. a store which only has the standard frames. This prevents snapshots from overwriting existing content in the store.

Using snapshots, we can now load the knowledge base store in less than 5 secs!

Bonus:
* Compare objects by value
* Fingerprint of sliced arrays
* Iterating all symbols in the symbol table
* Python methods for GC lock and unlock


tools/snaps.cc
```diff
@@ -0,0 +1,66 @@
...
+#include "sling/base/init.h"
+#include "sling/base/flags.h"
+#include "sling/base/logging.h"
+#include "sling/base/types.h"
+#include "sling/frame/serialization.h"
+#include "sling/frame/snapshot.h"
+#include "sling/frame/store.h"
+
+DEFINE_bool(check, false, "Check for valid snapshot");
+DEFINE_bool(verify, false, "Check snapshot by read it into memory");
```
anders-sandholm: Minor: read -> reading

ringgaard:
Thanks for the review.

PR #249, opened 2018-09-07
Category graph inversion

ringgaard:
I have added an extra task to the wiki processing pipeline for adding members of each category to the item for the category.

Extra:
* Add language for label and description for wikidata items
* Expose options to only use primary/known languages for item labels
* Expose options for pruning wikidata items


ringgaard:
Thanks for the review. You can use this for running through the member of the birth categories.

PR #251, opened 2018-09-07
Dynamic HTTP worker pool expansion

ringgaard:
The worker pool for the HTTP server is not expanded automatically when all the workers are busy.

This also fixes a bug in epoll_wait() error checking which could cause an infinite loop in the worker threads leading to the HTTP server becoming unresponsive.

I have also added support for plain text format in the web service handler.


ringgaard:
Thanks for the review

PR #252, opened 2018-09-10
Knowledge base snapshot

ringgaard:
Added --kb_snapshot flag to generate snapshot for knowledge base in wiki pipeline.


ringgaard:
Thanks for the review.

PR #254, opened 2018-09-10
PyArray slices

ringgaard:
I have implemented support for using slices of SLING arrays in Python. The slices support most operations without having to materialize the sliced array. You can even slice a slice.

I have also implemented support for compare-by-value for arrays and anonymous frames. This allows you to use anonymous frames and arrays as keys in Python dictionaries.

New isa() method for PyFrame.


sling/pyapi/pyarray.cc
```diff
@@ -116,8 +191,14 @@ int PyArray::Contains(PyObject *key) {
...
   ArrayDatum *arr = array();
-  for (int i = 0; i < arr->length(); ++i) {
-    if (arr->get(i) == handle) return true;
+  if (slice == nullptr) {
+    for (int i = 0; i < arr->length(); ++i) {
+      if (pystore->store->Equal(arr->get(i), handle)) return true;
+    }
+  } else {
+    for (int idx = slice->start; idx != slice->stop; idx += slice->step) {
+      if (pystore->store->Equal(arr->get(pos(idx)), handle)) return true;
```
rahul1980: do you need to use pos() here? The loop is from slice->start to slice->stop and not from 0 to length().

ringgaard:
Thanks for the review.

PR #259, opened 2018-09-17
Compact SLING documents

ringgaard:
I order to make the SLING document more compact and easier to read, I have shortened the names by removing the /s/document, /s/token, and /s/mention prefixes. 

I have also removed redundant frame slots, that can be inferred from the context:
* The /s/token/index is implied by the position in the index array.
* The :/s/token is implied by being in the token array.
* If /s/token/name is not specified, the token word is extracted from the document text.
* The default /s/token/break for the first token is now NO_BREAK.
* Default for token size is now 1 byte.
* The default mention type is now phrase

All the schemas that we no longer use from document-schema.sling has been removed.

The documentation has been updated to reflect the new document names.

The Document class now has a decorate() method to add the redundant document slots.

Support for adding extra slots when building document frames.

The token location functions have been consolidated into the Document class.

This change breaks compatibility with the old SLING document format. I have made a script for converting from version 1 to version 2 document format.


python/nlp/document.py
```diff
@@ -283,3 +286,20 @@ def refresh_annotations(self):
...
     self.themes_dirty = False
 
+
+class Corpus:
+  def __init__(self, filename, commons=None):
+    self.input = sling.RecordReader(filename)
+    self.iter = iter(self.input)
+    self.commons = sling.Store() if commons == None else commons
+    self.docschema = sling.DocumentSchema(self.commons)
+    if commons == None: self.commons.freeze()
```
rahul1980: no need to check if commons == None, since self.commons is being frozen here.

python/nlp/document.py
```diff
@@ -283,3 +286,20 @@ def refresh_annotations(self):
...
+
+  def __iter__(self):
+    return self
+
+  def next(self):
+    _, data = self.input.next()
```
rahul1980: don't you need to raise StopIteration if there is no more data.

tools/docv1to2.py
```diff
@@ -0,0 +1,62 @@
...
+{=token =/s/token}
+{=index =/s/token/index}
+{=start =/s/token/start}
+{=size =/s/token/length}
+{=break =/s/token/break}
+{=word =/s/token/text}
+{=phrase =/s/phrase}
+{=begin =/s/phrase/begin}
+{=length =/s/phrase/length}
+{=evokes =/s/phrase/evokes}
```
rahul1980: this doesn't convert /s/measure and /s/measure/domain

PR #260, opened 2018-09-18
Adding bot for storing fact records in wikidata (wikibot.py).

anders-sandholm:
Example of the outcome of a run can be found at https://www.wikidata.org/wiki/Q21621026
The output record file that is logging the work done in wikidata looks like this:
```
Q21621026: {=#1 
  facts: {=#2 
    P569: 1990
  }
  item: Q21621026
  status: {=#3 
    stored: "747706735"
  }
} 
```


python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+
+    self.source_claim = pywikibot.Claim(self.repo, "P3452") # Inferred from
+    self.time_claim = pywikibot.Claim(self.repo, "P813") # Referenced (on)
+    today = datetime.date.today()
+    time_target = pywikibot.WbTime(year=today.year,
+                                   month=today.month,
+                                   day=today.day)
+    self.time_claim.setTarget(time_target)
+
+  def getSources(self, category):
```
ringgaard: We usually use lowercase with underscores for method names, e.g. get_sources() instead of getSources().

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+      print item
+    print "Last record.", updated, "records updated."
+    self.status_file.close()
+    self.record_file.close()
+
+  def run(self):
+    self.storeRecords(self.record_file, batch_size=2)
+
+if __name__ == '__main__':
+  sfb = StoreFactsBot(test=True)
```
ringgaard: Should test be True here?

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+          self.n_item: fact_record[self.n_item],
+          self.n_facts: rs.frame({prop: val}),
+          self.n_status: rs.frame({self.n_stored: str(rev_id)})
+        })
+        self.status_file.write(item, status_record.data(binary=True))
+
+      updated += 1
+      print item
+    print "Last record.", updated, "records updated."
+    self.status_file.close()
```
ringgaard: This will fail if you call storeRecords multiple times.

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+        pass # unable to extract claims - move to next revision
+      else:
+        if prop in claims: return True
+    return False
+
+  def storeRecords(self, records, batch_size=3):
+    rs = sling.Store(self.store)
+    updated = 0
+    for item, record in records:
+      if updated >= batch_size:
```
ringgaard: What about being able to specify a range of records to update? 

You can add flags to control this for test runs, e.g.
```
import sling.flags as flags

flags.define("--first",
             help="first record to update",
             default=0)

flags.define("--last",
             help="last record to update",
             default=sys.maxint)

...
if recno < flags.arg.first or recno > flags.arg.last: continue

if __name__ == '__main__':
  flags.parse()
```

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+        print "Hit batch size of", batch_size
+        break
+      print "Processing", item
+      wd_item = pywikibot.ItemPage(self.repo, item)
+      if wd_item.isRedirectPage():
+        print item, " is a redirect page"
+        status_record = rs.frame({
```
ringgaard: What about having a method for logging errror/success to the status file?

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+        claim.setTarget(target)
+        cat_str = str(provenance[self.n_category])
+        summary = provenance[self.n_method] + " " + cat_str
+        wd_item.addClaim(claim, summary=summary)
+        claim.addSources(self.getSources(cat_str))
+        rev_id = wd_item.latest_revision_id
+        status_record = rs.frame({
+          self.n_item: fact_record[self.n_item],
+          self.n_facts: rs.frame({prop: val}),
+          self.n_status: rs.frame({self.n_stored: str(rev_id)})
```
ringgaard: Should we also store a timestamp of when it was updated?

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+    self.n_facts = self.store["facts"]
+    self.n_date_of_birth = self.store["P569"]
+    self.n_provenance = self.store["provenance"]
+    self.n_category = self.store["category"]
+    self.n_method = self.store["method"]
+    self.n_status = self.store["status"]
+    self.n_stored = self.store["stored"]
+    self.n_skipped = self.store["skipped"]
+    self.n_error = self.store["error"]
+    self.n_has_prop = self.store["already has property"]
```
ringgaard: I think it might be better to use strings for the error messages.

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,183 @@
...
+
+    self.store = sling.Store()
+    self.n_item = self.store["item"]
+    self.n_facts = self.store["facts"]
+    self.n_date_of_birth = self.store["P569"]
+    self.n_provenance = self.store["provenance"]
+    self.n_category = self.store["category"]
+    self.n_method = self.store["method"]
+    self.n_status = self.store["status"]
+    self.n_stored = self.store["stored"]
```
ringgaard: The n_stored is really the wikidata revision. What about calling it "revision" instead to make this explicit?

python/wikibot/wikibot.py
anders-sandholm: SG. Added "first" and "last" cmd-line flags as suggested.

python/wikibot/wikibot.py
anders-sandholm: Moved this boolean out so it is a cmd-line flag instead. HTH.

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,195 @@
...
+      record_file_name = "local/data/e/wikibot/birth-dates.rec"
+    status_file_name = "local/data/e/wikibot/wikibotlog-"+time_str+".rec"
+
+    self.record_file = sling.RecordReader(record_file_name)
+    self.status_file = sling.RecordWriter(status_file_name)
+
+    self.store = sling.Store()
+    self.n_item = self.store["item"]
+    self.n_facts = self.store["facts"]
+    self.n_date_of_birth = self.store["P569"]
```
ringgaard: This seems to be no longer used.

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,195 @@
...
+    self.n_category = self.store["category"]
+    self.n_method = self.store["method"]
+    self.n_status = self.store["status"]
+    self.n_revision = self.store["revision"]
+    self.n_url = self.store["url"]
+    self.n_skipped = self.store["skipped"]
+    self.store.freeze()
+    self.rs = sling.Store(self.store)
+
+    self.source_claim = pywikibot.Claim(self.repo, "P3452") # Inferred from
```
ringgaard: Inline comments are usually lowercase

python/wikibot/wikibot.py
```diff
@@ -0,0 +1,195 @@
...
+    self.status_file.close()
+    self.record_file.close()
+
+  def get_sources(self, category):
+    source_target = pywikibot.ItemPage(self.repo, category)
+    self.source_claim.setTarget(source_target)
+    return [self.source_claim, self.time_claim]
+
+  def ever_had_prop(self, wd_item, prop):
+    # up to 150 revisions covers the full history of 99% of e.g. human items
```
ringgaard: Standalone comments usually starts with uppercase

python/wikibot/wikibot.py
anders-sandholm: Good catch. Thx.

python/wikibot/wikibot.py
anders-sandholm: Fixed

python/wikibot/wikibot.py
anders-sandholm: Fixed

PR #261, opened 2018-09-19
Use compare by value semantics for anonymous frames in pyapi

ringgaard:
I have reverted back to comparing frames by reference in the Python API so that we can use anonymous frames as keys in dictionaries.


sling/pyapi/pyarray.cc
```diff
@@ -193,11 +194,11 @@ int PyArray::Contains(PyObject *key) {
   ArrayDatum *arr = array();
   if (slice == nullptr) {
     for (int idx = 0; idx < arr->length(); ++idx) {
-      if (pystore->store->Equal(arr->get(idx), handle)) return true;
+      if (pystore->store->Equal(arr->get(idx), handle), true) return true;
```
rahul1980: bracketing is odd. Similarly below.

sling/frame/store.h
```diff
@@ -989,14 +989,17 @@ class Store {
...
+  // reference.
+  bool Equal(Handle x, Handle y, bool byref = false) const;
 
   // Computes a fingerprint for an object. This fingerprint is independent of
   // the specific handle values in this store. Fingerprints of frames with ids
   // only depend on the name, no the content of the frame. The object cannot
-  // contain cycles.
-  uint64 Fingerprint(Handle handle, uint64 seed = 0) const;
+  // contain cycles. Anonymous frames can either be compared by value (default)
+  // or by reference. If they are by reference, the fingerprint will become
```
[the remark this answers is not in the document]
ringgaard: fixed

sling/pyapi/pyarray.cc
ringgaard: Oops! That's a bug. Fixed.

PR #262, opened 2018-09-19
Replace Collect and Lookup kernels with Gather

rahul1980:
Bonus: fix a corner case bug in ParserState, where we were erroneously disallowing any non-STOP action at the end of the sequence. This meant that any CONNECT actions at the very end would be disallowed.


sling/nlp/parser/trainer/pytorch_modules.py
```diff
@@ -640,21 +648,27 @@ def link(bldr, name, dim, cnx, prefix=True):
...
 
-      name = feature.name + "/Collect"
-      collect = ff.rawop(optype="Collect", name=name)
-      collect.add_input(indices)
-      collect.add_input(activations)
-      collected = ff.var(
-          name=name + ":0", shape=[feature.num, activations.shape[1] + 1])
-      collect.add_output(collected)
+      name = feature.name + "/Gather"
+      gather = ff.rawop(optype="Gather", name=name)
```
ringgaard: You could consider adding Gather to the FlowBuilder

sling/nlp/parser/trainer/pytorch_modules.py
```diff
@@ -640,21 +648,27 @@ def link(bldr, name, dim, cnx, prefix=True):
...
 
-      sz = feature.transform.weight.size()
+      projection = feature.transform.projection.weight
+      sz = projection.size()
       transform = ff.var(name=feature.name + "/transform", shape=[sz[0], sz[1]])
-      transform.data = feature.transform.weight.data.numpy()
+      transform.data = projection.data.numpy()
 
       name = feature.name + "/MatMul"
       matmul = ff.rawop(optype="MatMul", name=name)
```
ringgaard: Just curious... why do you use rawop for this MatMul instead of just using ff.matmul?

rahul1980:
Verified again on a larger corpus that it gives the same accuracy as the Myelin runtime.

PR #263, opened 2018-09-21
Fix bug in Wikipdia alias qid etc.

ringgaard:
- Fix bug in Wikipedia alias qid
- Add /w/item/member to schema
- Fall back to store if snapshot is invalid
- Fix qid for wikimedia disambiguation


ringgaard:
Thanks for the review.

PR #267, opened 2018-09-28
Add ability to process Wikibot logs

anders-sandholm:
Code added for processing log record files from running the Wikibot.


python/wikibot/dashboard.py
```diff
@@ -0,0 +1,114 @@
...
+precision_map = {
+  sling.MILLENNIUM: pywikibot.WbTime.PRECISION['millenia'],
+  sling.CENTURY: pywikibot.WbTime.PRECISION['century'],
+  sling.DECADE: pywikibot.WbTime.PRECISION['decade'],
+  sling.YEAR: pywikibot.WbTime.PRECISION['year'],
+  sling.MONTH: pywikibot.WbTime.PRECISION['month'],
+  sling.DAY: pywikibot.WbTime.PRECISION['day']
+}
+
+class Dashboard:
```
ringgaard: We usually use "dashboard" about something with a user interface. What about renaming this to something like WikiMonitor?

python/wikibot/dashboard.py
```diff
@@ -0,0 +1,114 @@
...
+    test = ""
+    if flags.arg.test: test = "test-"
+    self.pattern = self.path + "wikibotlog-" + test + "20*.rec"
+
+    self.store = sling.Store()
+    self.n_item = self.store["item"]
+    self.n_facts = self.store["facts"]
+    self.n_status = self.store["status"]
+    self.n_revision = self.store["revision"]
+    #self.n_url = self.store["url"]
```
ringgaard: pls delete this test line

python/wikibot/dashboard.py
anders-sandholm: Agree. Done

python/wikibot/dashboard.py
anders-sandholm: done

PR #277, opened 2018-10-10
Remove unused schemas

ringgaard:
I have removed the PropBank schemas since they are no longer being used by the new OntoNotes converter. I have also removed the VerbNet and WordNet schemas.


ringgaard:
We should go over the training documentation before the v2 release and update it to make sure that everything is aligned with the new pytorch parser trainer.

I thin we should simplify the whole commons business with the parser trainer. When you run `train_pytorch.py` we should create a new commons store. If the `--commons` flag is specified we load this into the store. Then we ALWAYS run over the train/dev (btw shouldn't dev be test?) corpora and add all the symbols needed. We then save this new store in the output folder and this store is the one that will end up in the commons data block in the final flow file.

The advantage of doing it this way is that you just need the train/dev corpora to train a parser model. You only need to use the `--commons` flags if you have additional schema data for the parser. With the new OntoNotes conversion (and this PR), we have effectively scrapped schema information for PropBank and OntoNotes, so the `--commons` never needs to be used with that corpus.

The downside is that you will have to recreate the commons file in the output directory every time you run the parser trainer, which takes 10 seconds. If this is annoying, you could have a `--reuse_commons` flag that skips the creation of the commons store file.

PR #278, opened 2018-10-11
Further tweaks to the Ontonotes Converter

rahul1980:
- Existing NP Expansion heuristic has been removed.
- Add extra noun phrases using base NP and NML constituents. This adds 173K and 3.5K spans respectively. Subsequent spans (e.g. arguments, coref) are normalized to these spans now, if possible.
- Trim trailing possessives (e.g. China's => China). This affects only ~700 spans.
- Particle inclusion only done for verb predicates now.
- Docid is stored as /ontonotes/docid instead of the frame id. This will allow 2 docs with the same docid to be read in the same store (e.g. gold and test doc).
- More head-finding rules for NML.
- Span generator method, which cleans up and simplifies span iteration code.
- Predicates are now /pb/<label> instead of /pb/pred/<label>. Arguments are now /pb/<label> instead of just <label>.

With all these changes, there are a total of 630K output spans now (up from 580K before this PR). The change is only +50K because we are able to align a lot more argument spans now.

Known minor issue: 724 argument spans still lie inside an NER span. All of them don't share the head with the NER. In 434 of these, the predicate of the argument is also inside the NER. This is because annotators have annotated inside NER spans as well. We have kept these 724 spans for now.


sling/nlp/parser/ontonotes/annotations.py
```diff
@@ -201,6 +208,26 @@ def __init__(self, converter):
     # Current sentence index.
     self.sentence = None
 
+
+  # Generator that returns spans of specified type(s).
+  def spans(self, types_mask, label=False):
+    if types_mask & NER > 0:
+      for s in self.ner.spans:
+        yield (s, "NER") if label else s
```
ringgaard: Now that you have defined constants for the span types (NER, PRED, ...), wouldn't it be easier to use these instead of the strings?

sling/nlp/parser/ontonotes/annotations.py
```diff
@@ -447,6 +478,99 @@ def _add_token_constituents(self):
       constituent.children = children
 
 
+  # Performs NP-expansion of 'span' beginning from its head.
+  # Expansion stop if it encounters a token that is disallowed.
+  # Returns the new span boundaries.
+  def _expand(self, span, disallowed):
+    assert span.head is not None
+
+    allowed_tags = ['NNS', 'NN', 'NNP', 'NNPS']
```
ringgaard: What about HYPH?

PR #292, opened 2018-11-16
Alias case forms and taxonomies

ringgaard:
Majority case forms for aliases in phrase table.

Type classification of items using taxonomy.

Entity popularity in knowledge base

Myelin compiler:
* Generate SVG files directly using DOT program.
* Dump generated assembly code directly using objdump program.
* Option for using Dragnn kernels in Myelin compiler.

Bonus:
* value() method moved to Date class so you don't need a Calendar object.
* Python API for frame evaluation.
* Fixed missing exception for type checking.
* UTF8 flag for frame serialization.
* Fix html table layout for task monitor UI.


sling/nlp/kb/facts.cc
```diff
@@ -78,6 +77,65 @@ void FactCatalog::Init(Store *store) {
...
+    "Q2198779",    // unit
+    "Q39875001",   // measure
+    "Q3695082",    // sign
+    "Q2996394",    // biological process
+    "Q838948",     // work of art
+    "Q47461344",   // written work
+    "Q28877",      // goods
+    "Q15401930",   // product
+    "Q483394",     // genre
+    "Q121769",     //  reference
```
rahul1980: extra space before 'reference'

PR #304, opened 2018-12-03
Use Python API for frame evaluation

rahul1980:
Bonus: Since the Python API doesn't need to read the common store from a file, we can get rid of commons_path at various places.


sling/nlp/parser/trainer/train_util.py
```diff
@@ -30,27 +30,16 @@ def check_present(args, ls):
...
-        ['bazel-bin/sling/nlp/parser/tools/evaluate-frames',
-         '--gold_documents=' + gold_corpus_path,
-         '--test_documents=' + test_corpus_path,
-         '--commons=' + commons_path],
-        stderr=subprocess.STDOUT)
-  except subprocess.CalledProcessError as e:
-    print("Evaluation failed: ", e.returncode, e.output)
-    return {'eval_metric': 0.0}
+def frame_evaluation(gold_corpus_path, test_corpus_path, commons):
+  metrics = sling.evaluate_frames(commons, gold_corpus_path, test_corpus_path)
```
ringgaard: You can now delete `import subprocess` on line 18.

On a related note, you also use psutil in this module, which is not a standard python module. You could consider using the standard `resource` module instead:

```
import resouce

def mem():
  info = resource.getrusage(resource.RUSAGE_SELF)
  return "(rss=%d)" % (info.ru_maxrss)
```

Then we could remove `sudo pip install psutil` from the [install guide](https://github.com/google/sling/blob/master/doc/guide/install.md).

rahul1980: Done.

PR #312, opened 2018-12-10
Matrix-matrix multiplication for integers

ringgaard:
* SIMD generator for integers using regular registers to support matrix-matrix multiplication.
* Add splits argument to Split op to make it compatible with tf and numpy.
* Added more tests


sling/myelin/simd-assembler.cc
```diff
@@ -1056,15 +1059,119 @@ class SSEScalarDoubleGenerator : public SIMDGenerator {
...
+    } else if (dst == src2) {
+      masm_->addq(reg(dst), reg(src1));
+    } else {
+      masm_->movq(reg(dst), reg(src1));
+      masm_->addq(reg(dst), reg(src2));
+    }
+  }
+
+  void Add(int dst, int src1, const jit::Operand &src2) override {
+    if (dst == src1) {
```
anders-sandholm: Seems like testing for the conjunction of dst == src1 and type_ == DT_INT64 in one go would make the code simpler because it would allow merging the two else branches

sling/myelin/simd-assembler.cc
```diff
@@ -1056,15 +1059,119 @@ class SSEScalarDoubleGenerator : public SIMDGenerator {
...
+        masm_->addq(reg(dst), reg(src1));
+      }
+    } else {
+      Load(dst, src2);
+      masm_->addq(reg(dst), reg(src1));
+    }
+  }
+
+  void Mul(int dst, int src1, const jit::Operand &src2) override {
+    if (dst == src1) {
```
anders-sandholm: Same comment as above (test conjunction and merge else branches)

sling/myelin/tests/myelin-vs-numpy.py
```diff
@@ -155,6 +162,9 @@ def simulate(flow, f, data):
       seq = []
       for k in range(n): seq.append(v[i[k]])
       v[o[0]] = np.concatenate(tuple(seq), axis)
+    elif op.type == "Split":
+      splits = np.split(v[i[0]], v[i[1]],v[i[2]])
```
anders-sandholm: nit: Add space after the second ","

sling/myelin/simd-assembler.cc
ringgaard: Yes. That's nicer. Done.

sling/myelin/simd-assembler.cc
ringgaard: Done.

sling/myelin/tests/myelin-vs-numpy.py
ringgaard: Fixed.

PR #315, opened 2018-12-17
Wikiflow uses latest dump files. Improved wikibot handling of date precision. 

anders-sandholm:


python/task/download.py
```diff
@@ -53,6 +55,8 @@ def run(self, task):
     # Download from url to file.
     if ratelimit > 0: log.info("Start download of " + url)
     conn = urllib2.urlopen(url)
+    srvLastModified = time.mktime(time.strptime(conn.headers['last-modified'],
```
ringgaard: The python convention is to use lowercase with underscores for variable names.

anders-sandholm: Thanks. Fixed.

PR #317, opened 2018-12-19
Extracting birth death dates from English Wikipedia articles

anders-sandholm:


python/wikibot/en_wp_dates.py
```diff
@@ -0,0 +1,141 @@
...
+                        int(match.group(1 + offset)))
+    else:
+      # MONTH DD, YYYY format"""
+      date = sling.Date(int(match.group(6 + offset)),
+                        self.months[match.group(4 + offset)],
+                        int(match.group(5 + offset)))
+    return date
+
+  def run(self):
+    month = "(January|February|March|April|May|June|July|August|September|"
```
ringgaard: What about "(" + self.months.values().join("|") + ")"

python/wikibot/en_wp_dates.py
```diff
@@ -0,0 +1,141 @@
...
+import sling
+import sling.flags as flags
+import re
+
+class ExtractWikipediaDates:
+  def __init__(self):
+    self.kb = sling.Store()
+    self.kb.lockgc()
+    print "loading kb"
+    self.kb.load("local/data/e/wiki/kb.sling", snapshot=True)
```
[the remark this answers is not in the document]
anders-sandholm: Cool. Removing it then.

python/wikibot/en_wp_dates.py
anders-sandholm: Yeah -- works if you do "|".join(months.keys()) instead. Fixed.

PR #324, opened 2019-01-14
Fix NotShiftOrMarkDelegate to work even if there is no MARK action

rahul1980:
When the training corpora only consists of single-token spans, MARK is not added to the action table.
Therefore actions.mark() will return None and break NotShiftOrMarkDelegate. This PR fixes the delegate's behavior in this corner case.


rahul1980:
Thanks for the quick review!

PR #326, opened 2019-01-21
Alias transfer

ringgaard:
We use anchors, redirects, and wikipedia titles as alias sources for building the phrase table, but these are all noisy sources. I have implemented "alias transfer" in the phrase table builder to clean up some of these noisy aliases. The basic idea is that an item cannot have the same name as another item that it is related to, except for a small list of exceptions (e.g. named after, different from etc.). The aliases are divided in reliable and unreliable aliases based on the source, and if an item has an unreliable alias and is related to another item which has the same alias from a reliable source, the alias count is transferred to the target.

The alias transfer procedure cleans up many cases of noisy aliases. For instance "botanist" is remove from "botany" because `[botanist (Q2374149) field of this occupation (P425): botany (Q441)]`.

Wikidata has finer granularity that any individual Wikipedia, so the alias transfer also fixes many cases where a "sub-item" has a redirect to a broader item, e.g. villages that are not in Wikipedia redirects to the county, taxon redirects to parent taxon etc. 

Removing all punctuation for alias normalization introduces too many false matches, so I have changed the default normalization for the phrase table so only dashes and periods are normalized. 

Other minor changes:

- Case form caching in tokens
- Fix rendering of thematic frames in docviewer
- Python methods for running workflows
- Fix strptime multi-thread problem
- Frame store handle overflow detection
- Fix memory leak for DocumentNames
- Move name and phrase tables to kb directory
- Support for mono-lingual text in WikiData
- Inverse calendar mappings
- Minor updates to default taxonomy
- Support for time period in fact extractor
- Use short names and country codes as aliases
- Handle "red links" in wiki annotator
- Stop using "intro text" as alias source since it is too noisy


sling/nlp/kb/phrase-table-builder.cc
```diff
@@ -0,0 +1,466 @@
...
+        num_instances_->Increment(count);
+      }
+    }
+  }
+
+  void Flush(task::Task *task) override {
+    // Prune phrase table by transfering unreliable aliases to reliable
+    // aliases for related items.
+    if (transfer_aliases_) {
+      LOG(INFO) << "Transfer aliases";
```
rahul1980: Do you still need the LOG(INFO) statements in this method?

sling/nlp/kb/phrase-table-builder.cc
```diff
@@ -0,0 +1,466 @@
...
+    int count() const { return count_and_flags & ((1 << 29) - 1); }
+    void set_count(uint32 count) {
+      count_and_flags = (count_and_flags & ~((1 << 29) - 1)) | count;
+    }
+
+    // Alias reliability.
+    bool reliable() const { return count_and_flags & (1 << 31); }
+
+    // Phrase form.
+    int form() const { return (count_and_flags << 29) & 3; }
```
rahul1980: shouldn't this be (count_and_flags >> 29) & 3 ?

sling/nlp/kb/phrase-table-builder.cc
ringgaard: I use it for verifying that alias transfer is enabled and it is also useful for timing the alias transfer.

sling/nlp/kb/phrase-table-builder.cc
ringgaard: Oops. That's a bug. Good catch!

sling/nlp/wiki/aliases.cc
```diff
@@ -66,18 +66,34 @@ class ProfileAliasExtractor : public task::FrameProcessor {
           AddAlias(&a, alias.GetHandle(n_name_), SRC_WIKIDATA_FOREIGN,
                    alias.GetHandle(n_lang_), alias.GetInt(n_count_));
         }
-      } else if (s.name == n_native_name_ || s.name == n_native_label_) {
+      } else if (s.name == n_native_name_ ||
+                 s.name == n_native_label_) {
```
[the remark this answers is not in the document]
ringgaard: Indeed it can.

sling/nlp/wiki/aliases.cc
```diff
@@ -238,7 +258,7 @@ class ProfileAliasReducer : public task::Reducer {
       merged.Add(n_alias_, a.Create());
     }
 
-    // Output alias profile.
+    // Output selected aliased.
```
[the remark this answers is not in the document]
ringgaard: Fixed

PR #336, opened 2019-02-25
Browser for category parses

rahul1980:
Allows browsing by category, signature, or top signatures (denoted by 'top').
Supports coarse and fine signatures, and three metrics to sort the parses with.
Allows customization of fact-matching scores.


sling/nlp/wikicat/browser.py
```diff
@@ -0,0 +1,889 @@
...
+    self._separator(header=False)
+    self.write_fact_match_counts(match_counts)
+    self._end("tr")
+    self._end("table")
+
+    # Write the individual parses in a tabular form.
+    self._br()
+    self._tag("b", "Categories with parses matching '" + signature + "'")
+    seen = set()
+    max_rows = 200
```
anders-sandholm: Probably fine with 200. Just wondering why 200? Might make sense to have it in a more central location if used elsewhere or simply in-lining it in the one place it is used. (Although the latter takes away the hint given by the name).

sling/nlp/wikicat/browser.py
```diff
@@ -0,0 +1,889 @@
...
+      selected = old_value(list_id, default_value)
+      for (name, value) in option_name_values:
+        if selected == value:
+          self._tag("option", name, value=value, selected="true")
+        else:
+          self._tag("option", name, value=value)
+      self._end("select")
+   
+    # Default weights for computing fact-match scores.
+    default_weights = {
```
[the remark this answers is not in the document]
rahul1980: Little bit of hand-tuning. My rationale was:
- We are interested in headroom, so NEW facts have a slightly higher weight. We are not super sure about ADDITIONAL, so they have a smaller weight.
- At the same time we want quality parses to score high, so it is good if a parse matches known facts (EXACT, SUBSUMES, SUBSUMED_BY).
- CONFLICT facts are bad, so they have a big -ve weight.

These are editable in the browser (and we can change the defaults in a follow up PR if needed).

sling/nlp/wikicat/browser.py
rahul1980: It is unlikely we will want to navigate very deep in the list so I chose 200. But we can change that (or implement a pagination mechanism to see the next 200).

PR #338, opened 2019-03-11
Token styles

ringgaard:
Document tokens can now have style attributes. This allows is to encode simple typographic markup in the documents. The style attribute on a token is an integer bit flag which encodes the start and end of the following styles:
- bold
- italic
- heading
- itemized lists
- block quotes

I have updated the text tokenizer to convert the html tags to token styles and the document viewer now also uses the token styles for rendering the documents with typographic markup.

I have made a simple LEX viewer with a UI that takes LEX-encoded input and displays the document using the document viewer. I have also updated the LEX encoder to output token styling.


sling/nlp/document/text-tokenizer.cc
```diff
@@ -213,14 +213,23 @@ BreakType TokenizerText::BreakLevel(int index) const {
...
 
   return NO_BREAK;
 }
 
+int TokenizerText::Style(int index) const {
+  TokenFlags flags = elements_[index].flags;
+  if (flags & TOKEN_TAG) {
+    TokenFlags style = flags & TOKEN_STYLE_MASK;
+    if (style != 0) {
+      return 1 << ((flags & TOKEN_STYLE_MASK) >> TOKEN_STYLE_SHIFT);
```
rahul1980: 1 << (style >> TOKEN_STYLE_SHIFT) ? (for simplicity)

sling/nlp/document/lex.cc
```diff
@@ -184,33 +212,52 @@ string ToLex(const Document &document) {
...
+          printer.PrintReference(frame);
         }
         output.WriteChar(']');
       }
     }
   }
 
+  // Terminate remaining styles.
+  if (styles != 0) {
+    OutputStyle(styles, &output);
```
rahul1980: is this correct? let's say there is an unterminated bold tag, so styles = BOLD_BEGIN, so OutputStyle() would erroneously open a <b> tag here, right?

sling/nlp/document/document.cc
```diff
@@ -469,7 +577,7 @@ Span *Document::EnclosingSpan(int begin, int end, bool *crossing) {
   for (int t = begin; t < end; ++t) {
     Span *s = tokens_[t].span_;
 
-    // Skip if is the same as the leaf span for the previous token.
+    // Skip if it is has the same leaf span as the previous token.
```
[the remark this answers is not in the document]
ringgaard: Fixed. Third time's a charm :)

sling/nlp/document/lex.cc
ringgaard: Maybe this is a little subtle, but the `styles` variable holds all the end tags that we have not seen yet. Line 219 (`styles |= begin_style << 1;`) uses the fact that the end tag mask is the begin tag mask shifted one to the left.

sling/nlp/document/text-tokenizer.cc
ringgaard: Done. Looks nicer.

sling/nlp/document/token-breaks.h
```diff
@@ -29,6 +29,67 @@ enum BreakType {
   CHAPTER_BREAK    = 6,
 };
 
+// Token style attributes.
+enum TokenStyles {
```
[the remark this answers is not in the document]
ringgaard: Yeah, token-breaks.h is a bit misleading now. Renamed.

PR #339, opened 2019-03-13
Infobox aliases and wiki links

ringgaard:
You can you specify in the infobox specification is a field should be used for alias extraction. 

This PR also includes a task for extracting a link graph from Wikipedia(s). This will be used in the next PR for entity resolution context scoring.


sling/nlp/wiki/aliases.cc
```diff
@@ -278,10 +284,12 @@ class AliasReducer : public task::Reducer {
       return !toxic;
     }
 
-    // Keep foreign, native and demonym aliases supported by Wikipedia aliases.
+    // Keep foreign, native and demonym, and nickname aliases supported by
```
rahul1980: nano nit: native and demonym -> native, demonym

sling/nlp/wiki/wikipedia-links.cc
```diff
@@ -0,0 +1,205 @@
...
+      link = store->Resolve(link);
+      if (store->IsFrame(link)) {
+        if (store->GetFrame(link)->IsNamed()) {
+          if (extract_mention_links_) {
+            (*links)[link]++;
+            num_theme_links_->Increment();
+            num_links_->Increment();
+          }
+        } else {
+          if (extract_infobox_links_) {
```
rahul1980: Is it guaranteed that all unnamed thematic frames are coming from infoboxes?
If there is an isa slot in 'f' that denotes it is coming from an infobox, maybe we should check for that slot here?

ringgaard: That's a good point. I have added a check for `isa:/wp/infobox`.

PR #340, opened 2019-03-13
Various Wikicat fixes

rahul1980:
- Omit outputting low-frequency (pid, qid) spans.
- Move subsumption checking code to Python
  - We only get the closure from C++ code now
  - Subsumption code in Python checks for genre, subclass, part_of, parent_org, located_in, and date subsumption. More properties can be added easily.
- Fix a minor bug in the browser.
- Expose the Resolve() method in the Python API.


rahul1980:
Thanks for the quick review!

PR #344, opened 2019-03-19
Minor fixes to the Wikicat browser

rahul1980:


rahul1980:
Thanks for the quick review!

PR #345, opened 2019-03-20
Fix subsumed calculation

anders-sandholm:


sling/nlp/wikicat/fact_matcher.py
```diff
@@ -229,9 +229,10 @@ def for_item(self, item, prop, value, store=None):
           return FactMatchType.EXACT
 
     # Check whether the proposed fact subsumes an existing fact.
-    closure_facts = self._existing_facts(store, item, prop, True)
-    if value in closure_facts:
-      return FactMatchType.SUBSUMES_EXISTING
+    for existing in exact_facts:
+      if isinstance(existing, sling.Frame):
+        if self.subsumes(store, prop[-1], value, existing):
+          return FactMatchType.SUBSUMES_EXISTING
```
rahul1980: Thanks for catching this.
You would also need to add a check for date-valued properties (like what we do below):
```
  if self._date_valued(prop[-1]):                                                                           
      for e in existing_dates:
        if self._finer_date(e, proposed_date):
          return FactMatchType.SUBSUMES_EXISTING
```

anders-sandholm: Good catch. Added.

PR #349, opened 2019-03-22
Fix refcounting bug in PyStore::Resolve()

rahul1980:


rahul1980:
Thanks for discussing this bug and confirming the solution Michael! Submitting this now, since you are already aware of the one line fix :)

ringgaard:
Thanks for fixing this.. Still LGTM.

PR #361, opened 2019-05-03
Tracking of entity handles in resolver

ringgaard:
I have added tracking of entity handles for anonymous frames to the entity resolver. This is preventing the locally generated frames from begin reclaimed by the garbage collector. I have also made the code more robust to missing entities. I can now run the NER labeling workflow without any crashes.

Bonus:
- Better memory usage reporting in task dashboard
- Record reader limit


run.sh
```diff
@@ -1,4 +1,4 @@
-#/bin/sh
+#!/bin/sh
```
anders-sandholm: Good catch!

PR #364, opened 2019-05-07
Rewrite Wikicat browser in AngularJS

rahul1980:
The existing browser was very unwieldy since it was based on form submissions, and was getting harder to extend. This PR rewrites the browser in AngularJS, and highly simplifies it.
The main changes are:
- AngularJS provides data binding for free (we were doing it ourselves in the old browser).
- We remove all kinds of scores, except the various match-type weights.
- We remove "coarse-signatures" and instead only keep full span signatures.
- We remove a few more settings, simplifying the look.
- We introduce a "parse selection formula" which allows the user to enter a python formula (with support for special placeholders) to restrict parses to those that pass the formula. 

Bonus: Dedup the set of examples for each match-type.


sling/nlp/wikicat/app/index.html
```diff
@@ -0,0 +1,380 @@
...
+          Top Signatures
+        </caption>
+        <thead>
+          <th>Index</th>
+          <th>Parse Signature</th>
+          <th>Example</th>
+          <th>Score</th>
+          <th>Selected Parses</th>
+          <th>Rejected Parses</th>
+          <th>Selected Match COunts</th>
```
anders-sandholm: minor nit: COunts -> counts

rahul1980: Done.

rahul1980:
> Thanks for rewriting the browser. Looking forward to trying it out.
> Maybe you did it already and I just didn't find it --- I was wondering if we could make the default behavior of the server such that it only serves/accepts local requests?

Added this check.

PR #367, opened 2019-05-10
Port Wikicat to Python3

rahul1980:
- Make 'print' a function
- Handle bytes <-> string conversions at a couple of places.

Bonus:
- Fix a sorting bug in Top Signatures
- Show an error message if our selection formula is too restrictive.
- Show a help message to clarify how signature scores are computed.


rahul1980:
Thanks for the review!

Issue #237, opened 2018-08-24
Clip gradient before optimizer update


ringgaard:
The gradient clipping in learning.cc should be applied to the incoming gradient (dv) instead of the final update to be compatible with TF and PyTorch. See #235.

PR #369, opened 2019-05-13
Refactor gradient clipping

ringgaard:
I have refactored gradient clipping as suggested in issue #237. This also allows be to consolidate the clipping code so it is implemented in the Optimizer class instead of having to implement this in each of the optimizers.


sling/myelin/learning.cc
```diff
@@ -107,6 +107,17 @@ void Optimizer::Build(Flow *flow) {
     gradmap[var] = tf.Ref(instance_[func], dvar);
   }
 
+  // Optionally add gradient clipping.
+  if (clipping_threshold_ != 0.0) {
+    auto *threshold = tf.Name(tf.Const(clipping_threshold_), "threshold");
+    for (auto &it : gradmap) {
+      auto &dv = it.second;
+      auto *norm = tf.Norm(dv);
```
rahul1980: FYI: This is different from how Pytorch implements clip_grad_norm_()   (which we use in the trainer):
https://pytorch.org/docs/stable/_modules/torch/nn/utils/clip_grad.html

There, the norm is computed over all parameters together (as if the gradients across all params were concatenated together). 

In Tensorflow too, while we have tf.clip_by_norm(tensor, clip_value) and tf.clip_by_global_norm(multiple_tensors, clip_value), the latter seems to be the more popular choice (akin to pytorch).

ringgaard: I have added support for both local and global gradient clipping. However, on the lextagger test, global clipping seems slightly less stable than local clipping, and the final model has ~0.2% lower accuracy.

PR #373, opened 2019-05-15
Port wikibot to python3.

anders-sandholm:
Port wikibot to python3 and add "--countonly" flag for wikimonitor.


python/wikibot/en_wp_dates.py
```diff
@@ -102,9 +102,9 @@ def run(self):
 
     for i in range(10):
       i_file = "local/data/e/wiki/en/documents-0000"+str(i)+"-of-00010.rec"
-      print i_file, records
-      for (item_id, record) in sling.RecordReader(i_file):
-        item = self.kb[item_id]
+      print(i_file, records)
+      for (item_id_bytes, record) in sling.RecordReader(i_file):
+        item = self.kb[item_id_bytes.decode()]
```
ringgaard: If you sync to the latest version, you can now use bytes for KB lookup.

anders-sandholm: done

PR #374, opened 2019-05-16
Cap nesting depth for MARK actions

rahul1980:
TODO: Max nesting depth is heuristically set to 5. we need to set this programmatically by looking at the training corpus.


rahul1980:
Thanks for the review.

PR #379, opened 2019-05-21
Fix str bug and add check to not upload previously deleted fact.

anders-sandholm:


python/wikibot/wikibot.py
```diff
@@ -188,19 +188,35 @@ def ever_had_prop(self, wd_item, prop):
...
+      else:
+        if prop in claims:
+          for claim in claims[prop]:
+            old_claim = pywikibot.Claim.fromJSON(self.repo, claim)
+            if old_claim.target_equals(target): return True
+    return False
+
   def log_status_skip(self, item, facts, error):
-    print("Skipping", str(item), " -- ", str(facts), error)
+    print("Skipping", str(item.id), " -- ", str(facts), error)
```
ringgaard: The print function will implicitly convert all arguments to string so the explicit str() conversions are not needed. You are welcome to postpone fixing this to a later PR.

anders-sandholm: Thanks. Will fix in a later PR

PR #382, opened 2019-05-24
Support for pagination and selection of individual categories in the Wikicat browser

rahul1980:
- Also increase top-k from 50 to 200.


sling/nlp/wikicat/app/wikicat.js
```diff
@@ -29,10 +30,15 @@ app.controller('SearchCtrl', function($scope, $http,  $location, $mdDialog) {
...
-  $scope.parse_selector = "";       // user-specified parse-selection formula
+  $scope.query = "";                   // main text box
+  $scope.query_results = null;         // results of the query
+  $scope.recordio_results = null;      // result of recordio generation
+  $scope.parse_selector = "";          // user-specified parse-selection formula
+  $scope.selected_parses = new Set();  // user-selected parses
+
+  // Current page and page size for displaying results.
+  $scope.current_page = 0;
+  $scope.page_size = 50;
```
anders-sandholm: Just wondering if it would make sense to use 200 here as well.

sling/nlp/wikicat/app/wikicat.js
```diff
@@ -160,15 +191,98 @@ app.controller('SearchCtrl', function($scope, $http,  $location, $mdDialog) {
...
   }
+
+  // Returns whether a particular row index should be displayed or not.
+  $scope.showTableRow = function(row) {
+    var start = $scope.current_page * $scope.page_size;
+    return (start <= row) && (row < start + $scope.page_size);
+  }
+
+  // Returns a [start, end) range list with the given step size.
+  $scope.range = function(start, end, step) {
```
anders-sandholm: JS not having standard libraries is just so awesome...

sling/nlp/wikicat/app/wikicat.js
rahul1980: I have increased it to 200, and the response size from the server to 500. I think we might hit the limits of AngularJS beyond that :-) For one, the recomputation cost after every UI change might be much bigger. Also, the size of the requests to the server might exceed the GET limits (~2K chars).
Please feel free to reduce this (and the server response size) if you face any problems.

sling/nlp/wikicat/app/wikicat.js
rahul1980: Yes, so annoying...

PR #383, opened 2019-05-28
Fix ontonotes conversion

rahul1980:
- Use id files from conll.cemantix.org/2012/coref instead of the current URL.
- Shuffle the training corpus using a fix (but configurable) random seed in the trainer.


sling/nlp/parser/tools/train_pytorch.py
```diff
@@ -32,7 +33,24 @@
 
 
 def train(args):
-  check_present(args, ["train_corpus", "output_folder", "dev_corpus"])
+  check_present(
+      args,
+      ["train_corpus", "output_folder", "dev_corpus", "train_shuffle_seed"])
+
+  train_corpus_path = args.train_corpus
+  if args.train_shuffle_seed > 0:
```
ringgaard: Wouldn't it be better to just build a shuffle index, and sample from this instead of having to write a new copy of the corpus to disk?

rahul1980: Yes, that is indeed what we discussed. The train corpus path is plumbed through to a few internal classes, so I will change that to take an index (or an iterator over docids) in a followup PR.

PR #388, opened 2019-06-13
Ensure qid label is a string in the parse signature

rahul1980:


sling/nlp/wikicat/prelim_ranker.py
```diff
@@ -150,11 +150,17 @@ def signature(self, document, parse, coarse=False):
         label = span.qid
       elif self.h_name in label:
         label = label[self.h_name]
+
+      # Ensure 'label' is a string.
+      if isinstance(label, sling.Frame):
+        label = "UNKNOWN" if label.id is None else label.id
```
anders-sandholm: Wondering if it would result in a better UX to use label.name or 'label.id + "(" + label.name + ")" ' rather than just 'label.id'

rahul1980: The signature is actually used as a key in various places. To get a better rendering of it as display text, we would need to make a few other changes elsewhere, not here.

rahul1980:
Thanks for the quick review!

PR #392, opened 2019-06-26
Fix dangling op producer

ringgaard:
Fix problem with variable producer pointing to non-existing op after elimination.


ringgaard:
Thanks for the review.

PR #397, opened 2019-07-08
Sparse embedding updates

ringgaard:
I have added support to Myelin for sparse tensors. A tensor can now have a sparsity tensor, which is a bitmap tensor indicating all non-zero sub-tensors relative to the first dimension. The SparseAdd kernel will request a sparse assignment target and update the sparsity bitmap for all the updates. The sparsity tensors are propagated through references, and the Calculate kernels in the optimizer can use the sparsity tensors for doing sparse updates of the global weights for the embedding vectors.

The sparse updates makes the optimizer in the factuality model 15 times faster (dense: 16701 us, sparse: 1084 us) so the optimizer only takes 1.19% of the CPU time compared to 15.56% with dense updates. Since the call to the optimizer are locked, this can become a bottleneck when doing multi-threaded training. The sparse updates make it much more scalable and allow us to scale up from 6 to 85 CPUs during training.


sling/myelin/express.cc
```diff
@@ -811,6 +811,143 @@ int Express::Complexity() const {
...
+bool Express::AlwaysZero(Var *x) const {
+  if (x->type == NUMBER && x->id == ZERO) return true;
+  Op *op = x->producer;
+  if (op == nullptr) return false;
+  switch (op->type) {
+    case MOV:
+      return AlwaysZero(op->args[0]);
+    case ADD:
+    case SUB:
+      return AlwaysZero(op->args[0]) && AlwaysZero(op->args[1]);
```
anders-sandholm: || (AlwaysOne(op->args[0]) && AlwaysOne(op->args[1]))  
would theoretically be correct although it would probably add nothing but overhead :-)

sling/myelin/express.cc
```diff
@@ -811,6 +811,143 @@ int Express::Complexity() const {
...
+    case MUL:
+      return AlwaysZero(op->args[0]) || AlwaysZero(op->args[1]);
+    case DIV:
+      return AlwaysZero(op->args[0]);
+    default:
```
anders-sandholm: I guess all the additional cases from ZeroFixpoint (e.g. case NEG: case ABS: , etc.) would fit in here as well? Or are they irrelevant/unlikely to any effect?

sling/myelin/express.cc
```diff
@@ -811,6 +811,143 @@ int Express::Complexity() const {
...
+      return AlwaysOne(op->args[0]);
+    case ADD:
+      return (AlwaysOne(op->args[0]) && AlwaysZero(op->args[1])) ||
+             (AlwaysZero(op->args[0]) && AlwaysOne(op->args[1]));
+    case SUB:
+      return AlwaysOne(op->args[0]) && AlwaysZero(op->args[1]);
+    case MUL:
+    case DIV:
+      return AlwaysOne(op->args[0]) && AlwaysOne(op->args[1]);
+    default:
```
anders-sandholm: case COS: 
  return AlwaysZero(op->args[0]);
;-)

sling/myelin/express.cc
```diff
@@ -811,6 +811,143 @@ int Express::Complexity() const {
...
+      return false;
```
anders-sandholm: Could these be added?
case ABS: case SQUARE: case SQRT: case FLOOR: case CEIL: case ROUND: case TRUNC:
  AlwaysOne(op->args[0])

sling/myelin/kernel/arithmetic.cc
```diff
@@ -243,25 +243,47 @@ struct Expression {
 
     // Compute the maximum common size between inputs and outputs. Scalars and
     // singular broadcasts are not used for computing the maximum size since
-    // these can be broadcast to the vector size.
+    // these can be broadcast to the vector size. Also, the expression is
+    // eligble for sparse computation if there is only one sparse non-trivial
```
anders-sandholm: eligble -> eligible

sling/myelin/express.cc
ringgaard: I have added this case as well.

sling/myelin/express.cc
ringgaard: While these are zero-fixpoints, i.e. -0=0 and |0|=0, they are not always zero for arbitrary inputs.

sling/myelin/express.cc
ringgaard: Hmm, cos(x)=0 is not true for all x.

sling/myelin/express.cc
ringgaard: Again, they are not always one.

sling/myelin/kernel/arithmetic.cc
ringgaard: Fixed

PR #400, opened 2019-07-17
Source and claim written in the same edit

vrandezo:
By adding the source to the claim before adding the claim to the item we only do one write to Wikidata (iiuc, not tested)

### DIFF
--- a/python/wikibot/wikibot.py
+++ b/python/wikibot/wikibot.py
@@ -374,10 +374,10 @@ class StoreFactsBot:
           continue
         summary = provenance[self.n_method] + " " + s
         claim.setTarget(target)
-        wd_item.addClaim(claim, summary=summary)
-        rev_id = str(wd_item.latest_revision_id)
         if len(sources) > 0:
           claim.addSources(sources)
+        wd_item.addClaim(claim, summary=summary)
+        rev_id = str(wd_item.latest_revision_id)
         self.log_status_stored(item, fact, rev_id)
         updated += 1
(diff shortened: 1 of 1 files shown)


vrandezo:
Just updating with my recent findings, I keep working on it (but wanted to have something here before heading out for the weekend).

Here's my notebook with the code that worked:

https://paws-public.wmflabs.org/paws-public/User:Denny/Testing%20Wikidata%20editing.ipynb

Here's the diff that worked out:

https://www.wikidata.org/w/index.php?title=Q4115189&diff=982087817&oldid=982056834

Maybe the pyWikibot version we are using is different?

PAWS has the following version installed: 
'[https] r-pywikibot-core.git (71e36c1, g10980, 2019/04/30, 16:42:46, OUTDATED)'

I'll try to run it locally to see if it makes a difference.

vrandezo:
Your script works on the desktop for me as well, both with Python 2 and 3.

Version:
Pywikibot: [https] wikimedia-pywikibot.git (0a74234, g11369, 2019/07/19, 12:12:16, ok)

Edit: https://www.wikidata.org/w/index.php?title=Q4115189&diff=982117100&oldid=982117056

anders-sandholm:
Thanks for looking into this again. Seems like we need to check and likely upgrade our version of pywikibot. I'm out for a few weeks but will look into this first thing when I'm back. Thanks again.

anders-sandholm:
Your suggested change works well using my now updated pywikibot installation. Thanks for your help (and patience).

PR #405, opened 2019-08-19
Fast snapshot GC

ringgaard:
The frame store snapshots made it fast to load large frames stores, but it was slow to freeze because the GC of the large global store. I have changed the snapshot format so it can now load the snapshot into "frozen" heaps. These are then read-only and all the objects have the mark bit set, so now GC takes less than 0.5 seconds.

I had to change the object layout in the frame store. One benefit of this is that a store can now have twice as many objects (512M to 1024M objects).

I have also implemented support for reading frames in RDF Turtle format, which we will need for Wikidata lexicographic data.


sling/frame/store.h
```diff
@@ -409,48 +404,53 @@ struct Range {
...
+//   |FTTM      payload size          | F=frame TT=type M=mark
 //   +--------------------------------+
-//   |      object handle          MTT| M=mark TT=tag
+//   |      object handle           TT| TT=tag
 //   +--------------------------------+
 //
+// The F bit indicates if the object is a frame. The TT bits for frames encodes
+// if the frame is a proxy and whether the frame has an id. The M bit is a mark
+// bit used for marking live objects during garbage collection. The mark bit is
+// also set for all object in frozen heaps.
```
anders-sandholm: object -> objects

ringgaard:
Thanks for the review. I have also added a better solution for the token type enum.

PR #409, opened 2019-08-29
Transition generation in C++

rahul1980:
Verified by running against the Ontonotes dev corpus (>9.6K sentences) and checking that we get the same actions as the Python implementation.

Bonus: Fix bug in frame/object.h, where we were holding a reference to a temporary functor.


sling/nlp/parser/trainer/generate-transitions.cc
```diff
@@ -0,0 +1,44 @@
...
+
+  DocumentSource *corpus = DocumentSource::Create(FLAGS_corpus);
+  sling::Store commons;
+  commons.Freeze();
+  for (int i = 0;; ++i) {
+    sling::Store store(&commons);
+    Document *document = corpus->Next(&store);
+    if (document == nullptr) break;
+
+//    std::cout << "Doc " << i << " text: "
```
ringgaard: Do you want to keep this debug comment?

sling/nlp/parser/trainer/generate-transitions.cc
```diff
@@ -0,0 +1,44 @@
...
+//              << document->PhraseText(0, document->num_tokens()) << "\n";
+    std::function<void(const ParserAction &)> callback =
+      [&](const ParserAction &action) {
+        std::cout << "Doc " << i << " " << action.ToString(&store) << "\n";
+      };
+    Generate(*document, callback);
```
ringgaard: I think it looks fancy to just include the lambda in the call:

```
Generate(*document, [&](const ParserAction &action) {
  std::cout << "Doc " << i << " " << action.ToString(&store) << "\n";
});
```

sling/nlp/parser/trainer/transition-generator.h
```diff
@@ -0,0 +1,37 @@
+#ifndef SLING_NLP_PARSER_TRAINER_TRANSITION_GENERATOR_H_
```
ringgaard: You are missing the copyright header here.

sling/nlp/parser/trainer/transition-generator.h
```diff
@@ -0,0 +1,37 @@
...
+
+namespace sling {                  
+namespace nlp {
+  
+// Generates transition sequences for [begin, end) token range in 'document',
+// calling 'callback' for every transition, and 'final_callback' at the end.
+void Generate(const Document &document,
+              int begin, int end,
+              std::function<void(const ParserAction &)> callback,
+              std::function<void()> final_callback);
```
ringgaard: What is the purpose of the "final" callback?

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+
+  // Information about a single frame.
+  struct Entry {
+    ~Entry() {
+      for (auto *edge : edges) delete edge;
+    }
+
+    Handle handle = Handle::nil();  // frame handle
+    Handle type = Handle::nil();    // type of the frame
+    bool span_based = false;        // whether the frame is evoked by text
```
ringgaard: You could consider using "evoked" instead of "span_based", i.e. a frame that is evoked from a span in contrast to a frame that is not evoked from a frame.

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+#include "sling/frame/object.h"
+#include "sling/frame/store.h"
+#include "sling/nlp/parser/trainer/transition-generator.h" 
+
+namespace sling {
+namespace nlp {
+namespace {
+
+// Map from a frame to information pertinent to transition generation.
+class FrameInfo {
```
ringgaard: Is it correct that "FrameInfo" is basically a frame graph where the nodes are Entry objects and edges are Edge objects? If so, you could consider using a terminology that is more graph-like, i.e. graphs, nodes, and edges.

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+
+  return output;
+}
+
+// Updates 'attention' as a result of execution 'action'.
+void Update(const Action &action, Handles *attention) {
+  auto type = action.core.type;
+  if (type == ParserAction::EVOKE || type == ParserAction::EMBED ||
+      type == ParserAction::ELABORATE) {
+    attention->insert(attention->begin(), action.frame->handle);
```
ringgaard: Wouldn't it be "cheaper" to reverse the order of the attention buffer so insertion and deletion is at the back?

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+  } else if (type == ParserAction::REFER || type == ParserAction::ASSIGN ||
+             type == ParserAction::CONNECT) {
+    std::remove(attention->begin(), attention->end(), action.frame->handle);
+    attention->insert(attention->begin(), action.frame->handle);
+  }
+}
+
+void InitInfo(
+    Frame frame, FrameInfo *frame_info, HandleSet *initialized) {
```
ringgaard: Do you need to have frame by value here? Could you use "const Frame &frame" instead?

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+        neighbor->edges.emplace_back(reverse);
+        reverse->inverse = edge;
+        edge->inverse = reverse;
+        pending.emplace_back(value);
+      }
+    }
+  }
+
+  for (Frame f : pending) {
+    InitInfo(f, frame_info, initialized);
```
ringgaard: What is the reason for not just calling InitInfo recursively inside the loop above?

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+    std::vector<TokenToMentions> *token_to_mentions) {
+  Handle h_evoked = document.store()->LookupExisting("evokes");
+  QCHECK(!h_evoked.IsNil());
+  HandleSet initialized;
+  token_to_mentions->resize(end - begin);
+
+  for (const Span *span : document.spans()) {
+    if (span->deleted() || span->begin() < begin || span->end() > end) continue;
+
+    for (const Slot &slot : span->mention().Slots(h_evoked)) {
```
ringgaard: It might be simpler to use span->AllEvoked() here.

sling/nlp/parser/trainer/transition-generator.cc
```diff
@@ -0,0 +1,457 @@
...
+    int index = span->begin() - begin;
+    if (span->length() == 1) {
+      (*token_to_mentions)[index].Singleton(span);
+    } else {
+      (*token_to_mentions)[index].Start(span);
+      (*token_to_mentions)[span->end() - 1 - begin].End(span);
+    }
+  }
+
+  for (Handle theme : document.themes()) {
```
ringgaard: Does this assume that all non-evoked frames are themes?

rahul1980: No. Any non-evoked frames reachable from evoked frames will be handled in InitNode() recursively.

sling/nlp/parser/trainer/transition-generator.h
rahul1980: Done.

sling/nlp/parser/trainer/transition-generator.h
rahul1980: To call a function after all the transitions have been generated.
For example, 'callback' can store transitions into a vector, and 'final_callback' can dispatch the vector to the trainer or a file writer.

sling/nlp/parser/trainer/generate-transitions.cc
rahul1980: Removed.

sling/nlp/parser/trainer/transition-generator.h
rahul1980: Done.

sling/nlp/parser/trainer/generate-transitions.cc
rahul1980: Done.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done. It does cause minor changes in the calling order though, so there are slight differences in the list of output transitions. But I have manually verified and all those diffs are legit.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done

rahul1980:
Thanks Michael, addressed all points.

sling/nlp/parser/trainer/transition-generator.cc
rahul1980: Done

rahul1980:
Verified by running against the Ontonotes dev corpus (>9.6K sentences) and checking that we get the same actions as the Python implementation.

Bonus: Fix bug in frame/object.h, where we were holding a reference to a temporary functor.

PR #411, opened 2019-09-11
Handle deletions from a local file

vrandezo:
Introduces a class that caches from a local file all relevant deleted triples and checks in that file first before it looks up in Wikidata, using the API, for more current possible deletions.

I didn't actually test it yet, I haven't come around to figure how to run it and test it, so it probably has errors. Feel free to ignore the PR until managed to run and test the code.

### DIFF
--- a/python/wikibot/wikibot.py
+++ b/python/wikibot/wikibot.py
@@ -16,6 +16,7 @@
 import pywikibot
 import sling
 import json
+import gzip
 import sys
 import datetime
 import sling.flags as flags
@@ -50,6 +51,11 @@ flags.define("--input",
              default="birth-dates",
              type=str)
 
+flags.define("--deletions",
+             help="gzipped file with all deleted triples",
+             default="deletions.triples.gz",
+             type=str)
+
 
 precision_map = {
   sling.MILLENNIUM: pywikibot.WbTime.PRECISION['millenia'],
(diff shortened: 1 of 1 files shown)


anders-sandholm:
If there's a location for the zip file, maybe I can try running it. Also, storing this in a SLING frame store will be faster and less memory consuming. But we can handle that in a later update.

python/wikibot/wikibot.py
```diff
@@ -60,6 +66,54 @@
...
+        if s not in self.deletions[p]: self.deletions[p][s] = []
+        self.deletions[p][s].append(o)
+      except ValueError:
+        pass
+
+  def contains(self, subject_id, property_id, object_value):
+    """Checks if a given triple was previously deleted."""
+    if self.path is None: return False
+    if property_id not in self.deletions:
+      self.load_deletions([property_id])
```
anders-sandholm: load_deletions -> load

python/wikibot/wikibot.py
```diff
@@ -197,8 +255,9 @@ def ever_had_prop(self, wd_item, prop):
     return False
 
   def ever_had(self, wd_item, prop, target):
-    # Up to 150 revisions covers the full history of 99% human items
-    revisions = wd_item.revisions(total=150, content=True)
+    in_deletions = self.deletions.contains(wd_item.title(), prop, target)
+    if in_deletions: return True
+    revisions = wd_item.revisions(starttime=self.deletions.timestamp(), content=True)
```
anders-sandholm: In this line I get the following error:
 pywikibot.data.api.APIError: badtimestamp_rvstart: Invalid value "<bound method Deletions.timestamp of <__main__.Deletions object at 0x7f09f7deefd0>>" for timestamp parameter "rvstart". [help:See https://www.wikidata.org/w/api.php for API usage. Subscribe to the mediawiki-api-announce mailing list at &lt;https://lists.wikimedia.org/mailman/listinfo/mediawiki-api-announce&gt; for notice of API deprecations and breaking changes.]
CRITICAL: Exiting due to uncaught exception <class 'pywikibot.data.api.APIError'>

anders-sandholm:
Closing this for now.

PR #414, opened 2019-10-07
Symbol table resizing

ringgaard:
Fix a bug in symbol tabel resizing for stores with frozen heaps. The resizing needs to take into account that all symbols from frozen heaps need to be last in the symbol hash bucket chains to make sure that symbols in the non-frozen heaps are properly handled by the garbage collector.


sling/frame/store.h
```diff
@@ -643,6 +643,14 @@ struct FrameDatum : public Datum {
     return false;
   }
 
+  // Checks if frame has slot with value.
```
anders-sandholm: Nit: // Checks if frame has a slot with name and value.

ringgaard: Fixed

PR #415, opened 2019-10-07
Fix a few wiki pipeline issues

ringgaard:
- Make demonyms language-dependent
- Fix bug in skip-table configuration
- Discard unaligned mentions
- Added more blacklisted aliases
- Fix problem with missing Japanese names


sling/nlp/wiki/wiki-annotator.cc
```diff
@@ -273,8 +273,18 @@ void WikiAnnotator::Category(const Node &node,
...
+    int b = a.begin.AsInt();
+    int e = a.end.AsInt();
+
+    // Map character span to token span. Discard annotation if span is not
+    // aligned with tokens.
+    int begin = document->Locate(b);
+    int end = document->Locate(e);
+    if (begin == end) continue;
+    if (document->token(begin).begin() != b) continue;
+
```
anders-sandholm: Just wondering if this could occur? 
if (document->token(end).end() != e) continue;

sling/nlp/wiki/wiki-annotator.cc
```diff
@@ -273,8 +273,18 @@ void WikiAnnotator::Category(const Node &node,
...
+    // Get character span for annotation.
+    int b = a.begin.AsInt();
+    int e = a.end.AsInt();
+
+    // Map character span to token span. Discard annotation if span is not
+    // aligned with tokens.
+    int begin = document->Locate(b);
+    int end = document->Locate(e);
+    if (begin == end) continue;
+    if (document->token(begin).begin() != b) continue;
```
anders-sandholm: Just a comment. Not sure how to fix this easily. Maybe if it's just me, but when you're not super familiar with the code it can be hard to tell token number/index ints from document/string/text positions/index ints.

sling/nlp/wiki/wiki-annotator.cc
ringgaard: It can in principle occur, but the Locate method will usually return a position that is to the right of the location, so the unaligned begin positions are more problematic than the unaligned end positions.

sling/nlp/wiki/wiki-annotator.cc
ringgaard: I tried to make it clear in the comment that we are mapping from character to token positions here. Luckily, we mostly deal with token positions in the remaining parts of the code.

PR #418, opened 2019-10-14
Document corpus reader

ringgaard:
- DocumentCorpus class with support for using a document schema

- Next() and Rewind() for RecordDatabase.

- Fix bug for empty spec in document annotator.

- Remove lex-viewer tool which has been replaced by the general analyzer tool.

- Support for feature vector padding in lexical encoder.

- Fix bugs in ParserState with untracked frames


sling/nlp/kb/phrase-table-builder.cc
```diff
@@ -284,6 +284,7 @@ class PhraseTableBuilder : public task::FrameProcessor {
         const EntityPhrase &e = phrase->entities[i];
         const Entity &entity = entity_table_[e.index];
         Handle item = commons_->Lookup(entity.id);
+        CHECK(!item.IsNil()) << entity.id;
```
anders-sandholm: Potentially also add 
```
        // Skip items that are not in the knowledge base.
        if (entity_item[source].IsNil()) continue;

```
inside the for-loop below.

ringgaard: We are trying to prevent that entity_item[source] is never nil, but we check this in the loop above, so this should never happen.

PR #421, opened 2019-10-31
Global gradient function registration and integer vector SIMD generators

ringgaard:
Gradient functions are now registered in a global registry instead of being part of a transformation library. This means that you no longer need to pass around the library from the compiler in order to build a flow for learning.

SIMD generators for 32 and 64 bit integers. Although I have implemented support for 16-bit integers in the accumulate and reduce code in the macro assembler there is no 16 bit vector SIMD generator since 16 bit masking is not supported.

I have integrated the Myelin code profiling into the general learner task

Rotating register allocator.

Support for read/write file mappings.


third_party/jit/assembler.cc
```diff
@@ -2239,15 +2249,25 @@ void Assembler::movq(Register dst, XMMRegister src) {
...
+  emit_sse_operand(src, dst);
+}
+
 void Assembler::movq(XMMRegister dst, XMMRegister src) {
   DCHECK(Enabled(MMX));
   EnsureSpace ensure_space(this);
   if (dst.low_bits() == 4) {
     // Avoid unnecessary SIB byte.
-    emit(0xf3);
+    emit(0xF3);
```
anders-sandholm: Seems there are other occurrences of lower case 'f' and 'e' other places in the code like line 2267 below. Would it makes sense to make this consistent once and for all or do you prefer slowly fixing them as we stumble on them going forward?

ringgaard: Thanks for the review. While I agree that it is a little annoying that the hex constants in the assembler are inconsistent w.r.t. case, I am a hesitating to just change them without having a good way of ensuring that we do not introduce new bug. These are notoriously hard to find.

PR #427, opened 2019-11-11
Parallel decompression of Wikidata dumps

ringgaard:
You can now use the --lbzip2 flag to enable parallel decompression of the Wikidata dumps using lbzip2. This speeds up the wiki-import stage from ~4 hours to ~1h.


python/task/wiki.py
```diff
@@ -275,7 +286,7 @@ def template_defs(self, language=None):
 
   def wikipedia_import(self, input, name=None):
     """Task for converting Wikipedia dump to SLING articles and redirects.
-    Returns article and redirect channels."""
+    Returns article, categories, and redirect channels."""
```
anders-sandholm: Good catch. Nice catch up.

python/task/workflow.py
```diff
@@ -541,6 +541,14 @@ def write(self, producer, output, sharding=None, name=None, params=None):
 
     return output
 
+  def pipe(self, command, format=None, name=None):
```
anders-sandholm: Awesome. Could very well come in handy in other situations as well as we move forward.

ringgaard: Thanks for the speedy review.

PR #436, opened 2019-12-20
Fix closing recordio files from pyapi

JDzvonik:
Hello there,
great project and sources! I really admire them.

I think I caught 1 bug. I've been using recordio pyapi with posix filesystem, as is. My use case included opening and closing recordio files in a loop and I quickly ran out of file descriptors on linux. I found a reason after investigation. Both `RecordReader` and `RecordWriter` classes should manage lifecycle of `File` pointer with already opened file. But `RecordReader` and `RecordWriter` instances created from pyapi received wrong argument, they opened their second instance of file and first file leaked. This fix worked for me.

So I'm sharing fix back.


JDzvonik:
@googlebot I signed it!

